PR #3, opened 2024-03-27
Skyler ic2 expansions

Skyler84:
Rebased this onto the new toolchain branch. 

Adds new python drivers for the I2C MUX and GPIO expander with APIs as close to machine.I2C/machine.Pin as is reasonable.

Please note while I've tested this on the linux micropython port I'm not able to completely confirm the machine interfaces and drivers are completely functional as it's not been tested on hardware. Anyone who is able to test - would be very appreciated if you could give it a try!

### DIFF
--- a/modules/tca9548a.py
+++ b/modules/tca9548a.py
@@ -1,0 +1,46 @@
+class TCA9548ABus:
+    def __init__(self, chip, bus_no):
+        self.chip = chip
+        self.bus_no = bus_no
+
+    def __getattr__(self, key):
+        chip = getattr(self, "chip")
+        bus_no = getattr(self, "bus_no")
+        try:
+            v = getattr(self, key)
+            return v
+        except:
+            v = getattr(chip.bus, key)
+            if callable(v):
+                def wrapper(*args, **kwargs):
+                    chip.select_downstream(bus_no)  
+                    return v(*args, **kwargs)
+                v = wrapper
+            setattr(self, key, v)
+            return v
+        
+class TCA9548A:
+
+    def __init__(self, bus, addr):
+        self.bus = bus
+        self.addr = addr
+        self.cur_i2c_bus = None
+        self.downstreams = [TCA9548ABus(self, i) for i in range(8)]
+        self.save = []
+
+    def select_downstream(self, bus) -> None:
+        if bus is not None and bus >= 0 and bus <=7 and self.cur_i2c_bus != bus:
+            self.system_i2c.writeto(self.addr,bytes([(0x1<<bus)]))
+            self.cur_i2c_bus=bus
+        elif self.cur_i2c_bus != None:
+            self.system_i2c.writeto(self.addr,bytes([0]))
+            self.cur_i2c_bus=bus
+
+    def save_downstream(self):
+        self.save.append(self.cur_i2c_bus)
+
+    def restore_downstream(self):
+        self.select_downstream(self.save.pop())
+
+    def get_downstream_bus(self, bus):
+        return self.downstreams[bus]
--- a/modules/tildagonos.py
+++ b/modules/tildagonos.py
@@ -1,6 +1,8 @@
 from machine import Pin, I2C, SPI
 import neopixel, time
 import gc9a01py as gc9a01
+from tca9548a import TCA9548A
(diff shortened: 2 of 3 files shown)


modules/tildagonos.py
```diff
@@ -27,78 +29,73 @@
...
         self.leds=neopixel.NeoPixel(Pin(21),19)
         self.gpiodata={}
         self.spi=None
         self.tft=None
+        self.pin_system_irq=Pin(10,Pin.IN)
+        self.pin_system_irq.irq(self.system_irq_handler, Pin.IRQ_RISING)
+
+        self.egpio_a = AW9523B(self.i2c_expansion.get_downstream_bus(BUS_SYSTEM), 0x58)
+        self.egpio_b = AW9523B(self.i2c_expansion.get_downstream_bus(BUS_SYSTEM), 0x59)
+        self.egpio_c = AW9523B(self.i2c_expansion.get_downstream_bus(BUS_SYSTEM), 0x5a)
```
b0xcat, line 48: self.init_gpio here seems to use the self.egpio instances declared below. Should it be moved down?

modules/tca9548a.py
```diff
@@ -0,0 +1,46 @@
...
+    def __init__(self, bus, addr):
+        self.bus = bus
+        self.addr = addr
+        self.cur_i2c_bus = None
+        self.downstreams = [TCA9548ABus(self, i) for i in range(8)]
+        self.save = []
+
+    def select_downstream(self, bus) -> None:
+        if bus is not None and bus >= 0 and bus <=7 and self.cur_i2c_bus != bus:
+            self.system_i2c.writeto(self.addr,bytes([(0x1<<bus)]))
```
b0xcat, line 33: self.system_i2c is not a member here, should it be self.bus?

modules/tca9548a.py
```diff
@@ -0,0 +1,46 @@
...
+            return v
+        except:
+            v = getattr(chip.bus, key)
+            if callable(v):
+                def wrapper(*args, **kwargs):
+                    chip.select_downstream(bus_no)  
+                    return v(*args, **kwargs)
+                v = wrapper
+            setattr(self, key, v)
+            return v
```
b0xcat, line 20: This seems to somehow recurse infinitely when used for writeto_mem from the egpio init calls

modules/tildagonos.py
```diff
@@ -27,78 +29,73 @@
...
-        self.system_i2c.writeto_mem(0x5a, 0x06, bytes([0xff,0xff]))
+        self.i2c_expansion.save_downstream()
+        sys_i2c = self.i2c_expansion.get_downstream_bus(BUS_SYSTEM)
+        self.egpio_a.init()
+        self.egpio_b.init()
+        self.egpio_c.init()
+        #I'd like this to not be direct access but it works as is.
         #chip C - set P0 bits 0,1,3,6,7 and P1 bits 0,1,2,3,4,5,6,7 to input
-        self.system_i2c.writeto_mem(0x5a, 0x04, bytes([0xcb,0xff]))
+        sys_i2c.writeto_mem(0x5a, 0x04, bytes([0xcb,0xff]))
```
b0xcat, line 67: After modifying some stuff to get here, this write seems to crash the board. I have no idea why, but possibly I just broke something in the code leading up to this :)

b0xcat (commented):
Hey Skyler, I tried running this and ran into some issues. I tried instantiating a tildagonos instance to test things with, if that's not the way to use this yet, my bad!

modules/tca9548a.py
b0xcat: Is this because v = wrapper also changes the v inside the wrapper causing it to call itself?

modules/tildagonos.py
Skyler84: ah yep, seems i missed that, thanks. i haven't been able to test this fully without hardware yet so testing appreciated!

modules/tca9548a.py
Skyler84: ahh, i thought i had that already hmmm

Skyler84:
This is no longer relevant as I'm re-writing everything in C

PR #5, opened 2024-03-30
Startup stuff

hairymnstr:
Added some board init stuff so that C functions (and therefore FreeRTOS tasks) can be run before Micropython is launched.


MatthewWilkes (approved):
Makes sense to me

PR #9, opened 2024-04-19
App scheduler / framework

b0xcat:
Hi! 

Here's an initial draft for a simple app scheduler.  Apps can be updated and drawn, and there is support for async background tasks. The idea is very similar to the [flow3r reactor / responder system](https://git.flow3r.garden/flow3r/flow3r-firmware/-/blob/main/python_payload/st3m/reactor.py?ref_type=heads), only in the current implementation apps can also be updated in the background. Maybe that's a bad idea though, depending on how slow things get.

I'm planning to build an event system too so apps can subscribe to events like button presses, sensor reads, or other types of input, but not sure what this is going to look like yet.

This is all pretty unfinished and subject to change, but if anyone has any thoughts or suggestions please let me know! I remember there being a discussion about using async or not to keep it accessible for beginning programmers, but for simple apps this is mostly hidden behind the sync `update` and `draw` methods. Hopefully that's ok!

Also, I'm not sure how loading apps from the app store / hexpansions will work yet, but I saw some code from previous badges that could definitely help there.


b0xcat:
Ok, I think this is in a reasonable state now! We can:

- Run apps in general, scheduled by the scheduler
- Start apps
- Stop apps
- Automatically launch apps from hexpansions
- Automatically stop hexpansion apps when they're unplugged
- Try to clean up modules imported from hexpansions when they're unplugged
- Have apps communicate through events, to keep them nice and decoupled

There's still a lot of things to be done like:
- Docs
- Telling hexpansion apps what hexpansion they're plugged in to, and probably giving them an object they can use for gpio access
- Integrating the egpio stuff from C instead of doing it from a python task with tildagonos

But this PR is getting kind of long, and this seems like a good base to build on top of in further PRs

PR #11, opened 2024-04-25
Simulator support

MatthewWilkes:
@b0xcat This makes the simulator work with your branch, and includes the async hardening discussed


b0xcat:
I think I found a slightly simpler way of getting CPython compatibility, how is this?

b0xcat (commented):
Oops, I didn't know double underscore variable names got mangled!

PR #12, opened 2024-04-28
DRAFT: async app framework tweaks

MatthewWilkes:


MatthewWilkes:
This is a WIP. General changes:
1. Rename some functions to be clearer about what they do
2. Switch the update/background split to be about background/foreground not sync/async
3. Introduce run method that's similar to the background method, and a background update similar to update, where there's a default impletnation of the async loop that calls the sync method. This lets users choose the sync update style if they want something simple, or reimplement the loop if they want to be more procedural
4. Make the yes/no dialog awaitable, rather than requiring callbacks
5. Add missing __init__.pys
6. Add a sample basic app demoing the above
7. Alter the way the locking works in the scheduler to be a bit more decoupled, to allow the async approach

MatthewWilkes:
@b0xcat What do you think about this? The `update(delta)` option is retained for more event-driven apps. The basic app feels easier to me than the equivalent event-driven one.

modules/apps/basic_app.py
```diff
@@ -0,0 +1,39 @@
...
+
+from app_components.dialog import YesNoDialog
+
+
+class BasicApp(app.App):
+    def __init__(self):
+        super().__init__()
+        self.color = (1, 0, 0)
+
+    async def run(self, render_update):
```
MatthewWilkes, line 14: @b0xcat In particular, this method for writing simple apps I think is good, and clearer than using background update.

modules/frontboards/twentyfour.py
```diff
@@ -0,0 +1,43 @@
...
+    BUTTON_PINS = {
+        BUTTONS["A"]: (0x5A, 0, (1 << 6)),
+        BUTTONS["B"]: (0x5A, 0, (1 << 7)),
+        BUTTONS["C"]: (0x59, 0, (1 << 0)),
+        BUTTONS["D"]: (0x59, 0, (1 << 1)),
+        BUTTONS["E"]: (0x59, 0, (1 << 2)),
+        BUTTONS["F"]: (0x59, 0, (1 << 3)),
+    }
+
+    async def background_task(self):
```
MatthewWilkes, line 31: ... and this background task allows app authors to be explicit about what code is needed always and what when the app is running the badge

b0xcat:
I haven't had time to take a proper look yet, will do asap!

MatthewWilkes:
No worries, was just marking it as ready tbh. Had some extraneous changes
in previously

modules/app.py
```diff
@@ -1,14 +1,45 @@
 import asyncio
+import time
+
 
 class App:
     def __init__(self):
+        self.overlays = []
         pass
 
+    async def run(self, render_update):
```
b0xcat, line 10: My first reaction to this was kind of like, it feels like we're pushing part of the job of the scheduler onto the app here, but I think it's ok since it's kind of hidden in the app base class.

This may be more my personal bias, but I am a bit worried about using inheritance since it makes the logic a bit harder to follow imo, but if we don't overuse it it should be ok.

modules/apps/basic_app.py
b0xcat: It does allow a much more linear flow of the app logic without blocking thanks to async, so I definitely agree this is a nice way to do things. 

I tried to stick to sync code as much as possible to keep things accessible for beginning programmers, but like this they can still do that if they want which is great.

modules/system/scheduler/__init__.py
```diff
@@ -124,34 +135,55 @@ async def _handle_request_foreground_pop(self, event):
...
+        async def mark_update_finished():
+            # Unblock renderer
+            self.render_needed.set()
+            await asyncio.sleep(0.001)
+
+            # If we're no longer foregounded, wait until it is before returning
+            did_lose_focus = False
+            while not self.app_is_foregrounded(app):
+                did_lose_focus = True
+                await asyncio.sleep(100)
```
b0xcat, line 159: This sleep seems a bit long, is that intentional?

b0xcat: If this is ok, then feel free to merge!

MatthewWilkes: It wasn't, meant to do millis or micros I think. I'm away for a day, wanna fix this and merge both of ours down to main?

modules/app.py
MatthewWilkes: Yeah, that's a fair comment. I think the base class version is quite different to what people will write if they want to override this, at least? Those will be more linear

modules/system/scheduler/__init__.py
b0xcat: Yeah sounds good!

modules/app.py
b0xcat: Yeah exactly!

PR #14, opened 2024-05-04
Add menu widget

hughrawlinson:
Maybe also needs methods to modify the menu items and set the current position externally?


MatthewWilkes:
Love it, works well for me.

PR #22, opened 2024-05-06
Add basic app launcher

MatthewWilkes:


Molive-0:
It seems that the launcher is not in the branch?

`ModuleNotFoundError: No module named 'system.launcher`

MatthewWilkes:
Oops, thanks @Molive-0. Added now.

modules/system/launcher/app.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def launch(self, item):
+        module_name = item["path"]
+        fn = item["callable"]
+        app_id = f"{module_name}.{fn}"
+        app = self._apps.get(app_id)
+        print(self._apps)
+        if app is None:
+            print(f"Creating app {app_id}...")
+            module = __import__(module_name, None, None, (fn,))
```
nyaalex, line 118: Currently this will try to import the app name as-is. However, since the apps are all in the `/apps/` folder and that is not in the `sys.path` this will fail. This should be replaced with `__import__('apps.' + module_name, [...]` or similar. Alternatively you can append `/apps/` to the `sys.path`.

MatthewWilkes: Good catch. Thanks, I was just coming in to look at this. I assumed it had been a simulator path issue.

nyaalex: No problem! Glad I could help :grin:

MatthewWilkes:
@nyaalex @Molive-0 I've pushed a quick commit with an example app that loads correctly in the simulator. This is still in active dev, so can't promise it won't change more, but this should be enough for people to try stuff more easily.

MatthewWilkes:
Going to merge this for now. Think it's a step forward, and want to get the simulator out.

PR #23, opened 2024-05-06
Add emf brand font

hughrawlinson:
From [here](https://www.emfcamp.org/about/branding)

The font is using the open font license, but the flow3r folks have taken the name clause to mean that they need to rename their version of the font - that's why this is called EMFCampHeading/EMFCampLabel.

It compiles fine, but when I try to use ctx.get_font_name to get the list of fonts, it just gives the existing list (including the flow3r fonts which we may want to remove?)

Here's how to list the fonts in the ctx:

```python
def list_fonts(ctx):
    font_names = []
    i = 0
    while True:
        try:
            font_name = ctx.get_font_name(i)
            font_names.append(font_name)
            i += 1
        except IndexError:
            break
    return font_names
```


thinkl33t:
My reading of the OFL is that since we're embedding it in our software, as long as we ship the license and don't modify the *shape* of the font, its still Raleway and can be called such (based on the OFL FAQ @  https://openfontlicense.org/how-to-use-ofl-fonts/ )

hodefoting:
@thinkl33t see https://openfontlicense.org/how-to-modify-ofl-fonts/ "5.9  Do font rebuilds require a name change? Do I have to change the name of the font when my packaging workflow includes a full rebuild from source?"

The re-encoding of an OpenType font as a "native ctx-font" is a rebuild. Coordinates of the control points of shapes get re-quantized, the glyph coverage is subsetted, most ligatures are dropped; and many advanced more advanced OpenType features are dropped.

PR #35, opened 2024-05-18
Basic OTA support

MatthewWilkes:
This ports the OTA app and helpers from tidal, missing most of the user interactivity, but allowing updates


MatthewWilkes:
Given we're close to the event, I'm going to merge this. Happy to revert or change based on comments, there's certainly things that need to change.

PR #38, opened 2024-05-20
Improve ctx with fixed alpha channel

npentrel:
I fixed the rgba method to have values from 0-1.0 for alpha channel (as I think is customary) in the ctx library and added the error handling that there was a todo for.

### DIFF
--- a/sim/fakes/ctx.py
+++ b/sim/fakes/ctx.py
@@ -304,20 +304,55 @@ class Context:
         self._emit(f"gray {v:.3f}")
         return self
 
+    def _value_range_rgb(self, value, value_str, low_limit, high_limit):
+        if value > high_limit:
+            print(
+                "{name} value should be below {limit}, this is an error in the real uctx library. Setting to {limit}.".format(name=value_str, limit=high_limit),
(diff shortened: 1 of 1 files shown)


MatthewWilkes:
This looks good, but I'm not convinced that the comment here is correct, trying on an actual badge I'm having trouble getting correct behaviour for the 0..256 range for r,g,b

npentrel:
Sorry which comment do you mean?

npentrel:
I have only tested with the simulator where this does seem to work for me. What behaviour are you seeing?

MatthewWilkes:
I meant the todo in the simulator code. It talks about automatically dispatching depending on the 0..255 or 0..1 ranges, if I'm understanding correctly. However, on the physical badge I was seeing only the 0..1 range working right. Let me check this again.

npentrel:
Is that in `_sim.py`? I only see a TODO there about LEDs which is unrelated I think

MatthewWilkes:
I'd meant the todo you referenced, but I double checked and it's working fine. I'm not sure what had confused me before, sorry for the delay merging.

PR #40, opened 2024-05-22
Add menu item change handler

emericklaw:
Allows code to be executed when the menu item is being changed.

My use case is to preview LED effects without having to press the confirm button and show a nice preview of the colour palette on the menu.

VirtualBoxVM_2024-05-22_06-06-50


MatthewWilkes:
Thank you for the contribution, and especially the demo screenshot!

PR #41, opened 2024-05-23
Text Dialog

jake-walker:
I'm messing around making an app and needed a way to get text entry, so I made a little text dialog based off the yes/no dialog.

It has upper and lowercase letters which are cycled through with the up and down buttons. The right button adds the character to a buffer, left removes the last character. Then confirm exits the dialog, setting the result to the buffer and setting cancel just closes down the dialog. There's also a `masked` option for password entry.

I've got no real hardware to run this on, but it works in the simulator. Here's a small demo:

Text Dialog Demo


MatthewWilkes:
Thank you for contributing this!

PR #42, opened 2024-05-24
tildagon_i2c: set SCL freq to 133KHz (needed for installation at EMF)

Skyler84:
Turns out *some* devices ~~cough ATSHA204A~~ won't exit from sleep at 400KHz clock. This ensures compatibility with the aforementioned chip which is used in a badge-enabled installation at camp this year (and probably in the future too).


ChrisDick:
I've tested the usb-pd devices at the lower speed and they're ok.

MatthewWilkes:
Thanks for testing that, @ChrisDick! Going to merge.

PR #44, opened 2024-05-24
pattern gen pull from settings

thinkl33t:


modules/patterns/off.py
```diff
@@ -0,0 +1,8 @@
+from patterns.base import BasePattern
+
+
+class OffPattern(BasePattern):
+    def __init__(self):
+        super().__init__()
+        self.fps = 0
```
MatthewWilkes, line 7: This will trigger a divide by zero at ` await asyncio.sleep(1 / self._p.fps)`

MatthewWilkes:
LGTM other than the 0 fps in the off pattern.

modules/patterns/off.py
thinkl33t: I intended that that line will never run if the fps is set to 0, as it breaks out of the `while: True` and stops execution.

PR #48, opened 2024-05-26
Add support to mirror the pattern colour onto the hexpansion detect LEDS

thinkl33t:


MatthewWilkes (approved):
Looks sensible to me.

PR #51, opened 2024-05-27
App error handling

MatthewWilkes:
Adds checks for tasks raising exceptions then triggers a stop app event to cleanly shut the app down.

Also:


thinkl33t:
approving on strength of the easter egg

PR #60, opened 2024-05-29
Add ujson fake

jake-walker:
This adds in a very simple mock for `ujson` which I think is a built-in MicroPython library for which mocks are missing.


MatthewWilkes:
Fine to add this, but I'd recommend using `import json` as the `ujson` name is a backwards compatibility alias.

PR #68, opened 2024-05-30
HMAC: Add tildagon HMAC module

Skyler84:
Add tildagon.HMAC module. Has been tested and verified working with burned HMAC key using burn_mac.py.

Keys should be provisioned on KEY0 block with HMAC purpose (handled automatically by burn_mac.py)

### DIFF
--- a/burn_mac.py
+++ b/burn_mac.py
@@ -1,0 +1,46 @@
+#!/usr/bin/python
+
+import hashlib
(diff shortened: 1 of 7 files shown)


Skyler84:
@MatthewWilkes could we get this merged with a release please, Slightly need it for GCHQ.NET

Issue #75, opened 2024-05-31
M24C16-RMN6TP in approved EEPROM list, but addressing appears incorrect


gsuberland:
The M24C16-RMN6TP is in the list of approved EEPROMs [here](https://tildagon.badge.emfcamp.org/hexpansions/creating-hexpansions/).

The [datasheet](https://www.st.com/en/memories/m24c16-r.html) shows that the M24C16-R series of EEPROMs uses the address format `0b1010xxxx`, with the bottom bits used for page select and R/W.

![image](https://github.com/emfcamp/badge-2024-software/assets/6430604/8ccdc31e-984b-428d-9afc-b002576d5a6e)

This means that the base address of this EEPROM series is **0xA0**.

However, the badge firmware uses 0x50 or 0x57 for addressing EEPROMs:

https://github.com/emfcamp/badge-2024-software/blob/e8d716ad249e77833d2ef2865fb29c304a85203c/modules/system/hexpansion/util.py#L9-L15

As such, unless I am missing something critical, it is not possible for any hexpansions which use the M24C16-R series EEPROMs to be detected and read from with the current firmware.

BartDeWaal:
The rightmost bit is R/W, so it doesn't count for the address. So you need to shift the address right by 1.

That means the address is 1010000, so 0x50. This is correct.

hairymnstr:
There are certainly issues with this EEPROM which need addressing, see #129. However, I am certain that addressing is not the issue so I'm going to close this particular issue.

gsuberland:
No worries, it does seem that the address thing was a red herring.

PR #76, opened 2024-05-31
Add 0xA0 to EEPROM address detection for M24C16-R

gsuberland:
The docs list M24C16-RMN6TP as a supported EEPROM for hexpansions, but the M24C16-R series uses addresses 0xA0 to 0xA8 for its page addressing. The code currently only supports 0x50 and 0x57 for addresses, as discussed in #75. This PR adds support for 0xA0 as a base address.

Note: **THIS HAS NOT BEEN TESTED.** I don't currently have time to spin up an environment for badge firmware development. I'm providing this in the hopes of speeding up an OTA rollout so we can get the nullsector hexpansions programmed and running, since we used those EEPROMs. Given that the addressing is wrong, it is unclear if the M24C16-R series has other behaviours that have not been tested with the badge. If you need me to provide one of our hexpansions for testing you can reach me via DECT on 6464 and I can deliver one.

### DIFF
--- a/modules/system/hexpansion/util.py
+++ b/modules/system/hexpansion/util.py
@@ -10,6 +10,8 @@ def detect_eeprom_addr(i2c):
     devices = i2c.scan()
     if 0x57 in devices and 0x50 not in devices:
         return 0x57
+    if 0xA0 in devices and 0x50 not in devices:
+        return 0xA0
     if 0x50 in devices:
         return 0x50
     return None


hairymnstr:
I'm working on flopagon EEPROM which uses this, addressing is not the issue (0x50 << 1 = 0xA0, the shift is done inside the I2C driver).  I can write headers but the littlefs filesystem is a problem.

gsuberland:
Interesting. I have a bit of C# that I wrote to generate a minimal EEPROM image, so perhaps that might help you? I can post it up in a repo if you want. Should be fully cross platform since it targets .NET 7/8.

gsuberland:
Closing this as the issues with this EEPROM are not simply down to addressing, as per #129.

PR #82, opened 2024-05-31
Eeprom stuff

hairymnstr:
Changes to make EEPROMs work.

* Changed a bunch of functions to use addr_len passed in as an explicit indicator of address length for the EEPROM.  Previously it was just writing two bytes and then trying one if it got an error.  That would only throw if your EEPROM was write protected! Otherwise it was writing a byte to the EEPROM.
* Changed the default block size for littlefs from 512bytes to 64bytes.  On small EEPROMs (e.g. 2048 bytes) you'd only get three blocks once metadata was removed which isn't enough for a root directory and a file of any useful size.

Needs testing on 16bit addressed EEPROMS


eclispe:
Was just looking at fixing some of these myself, looks like [write_header](https://github.com/emfcamp/badge-2024-software/blob/6b2c29ee42841f61c28a81dfbdf0bb81fba11d62/modules/system/hexpansion/header.py#L97C1-L97C17) also needs its address length fixing

hairymnstr:
Should address all the issues with the M24C16 EEPROM.

* Leaves block size at 512 bytes for the Zetta EEPROM ( so deployed ones aren't broken)
* Fixes some dodgy identification of non-EEPROM devices (above 0x57) as non-contiguous EEPROMs
* Fixes the write_header function so it uses the right address length

Tested on:
* Matthew Wilkes' Mega Drive Hexpansion (ZD24C64A)
* Danny Walker's Caffeine Buzz Hexpansion (ZD24C64A)
* My own Flopagon Hexpansion (M24C16)

MatthewWilkes:
Tested on Googly Eye and Flopagon, LGTM also.

PR #93, opened 2024-06-01
Fix app installation on the simulator

nomis:
Python's tarfile doesn't append a "/" to directory names like MicroPython does, so the root directory detection fails. Normalise the names by removing the "/" and then adding it back when returning the root directory name.


nomis:
I've only tested this on the simulator and the app store appears to be down now, hence the handling of the 404 response from the demo server.

thinkl33t:
The app store is now working again, if you could have a test of this and let u know if its working it'd be appreciated :)

nomis:
I've tested this on the badge and it works

PR #101, opened 2024-06-03
Add step counter to IMU API

Gregox273:
Enables the step counter in the BMI270, exposing a `step_counter_read` method as part of the IMU API.

Example usage:
```python
import app
import imu

class ExampleApp(app.App):
    def __init__(self):
        self.acc_read = None
        self.steps_read = None

    def update(self, delta):
            self.button_states.clear()
            self.minimise()
        else:
            self.steps_read = imu.step_counter_read()

    def draw(self, ctx):
        ctx.save()
        ctx.rgb(0.2,0,0).rectangle(-120,-120,240,240).fill()
        if self.steps_read:
            ctx.rgb(1,0,0).move_to(-80,-40).text("steps:\n{}\n".format(self.steps_read))
        else:
            ctx.rgb(1,0,0).move_to(-80,0).text("no readings yet")
        ctx.restore()

__app_export__ = ExampleApp
```

### DIFF
--- a/components/flow3r_bsp/flow3r_bsp_imu.c
+++ b/components/flow3r_bsp/flow3r_bsp_imu.c
@@ -23,10 +23,12 @@ static const char *TAG = "flow3r-imu";
 static void bmi2_error_codes_print_result(int8_t rslt);
 static int8_t set_accel_config(flow3r_bsp_imu_t *imu);
 static int8_t set_gyro_config(flow3r_bsp_imu_t *imu);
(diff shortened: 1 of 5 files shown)


ChrisDick:
Thanks for this, it was on my to do list. I've given this a try and it is working for me. Can you add a stub for the sim in sim/fakes/imu.py, then I'll merge it.

MatthewWilkes:
Thank you for this, @Gregox273! I've added a simple stub to the simulator and merged this.

PR #103, opened 2024-06-05
Add support for setting egpio output pin mode (and in)

emericklaw:


emericklaw:
An example app that uses this method of usign output egpio pins can be found here: https://github.com/emericklaw/tildagon-app-pacman-led

Robotmad:
Looks good to me and needed asap as workaround encourages use of direct access to I2C(7).

PR #108, opened 2024-06-05
Improve bootloader activation instructions

futureshape:
Bootloader doesn't work if the device is already powered on when plugging in USB


ChrisDick:
Thanks! We'd missed that step.

PR #110, opened 2024-06-05
Update hexpansion button chord to also control the LEDs

emericklaw:
To prevent duplication of code I moved the LED control inside the event, this might not the the right way of doing things...

Also changed the logic to allow you to use the same chord to remove a hexpansion.

### DIFF
--- a/modules/frontboards/twentyfour.py
+++ b/modules/frontboards/twentyfour.py
@@ -34,7 +34,14 @@ class TwentyTwentyFour(FrontBoard):
         display.gfx_init()
 
         button_states = {button: False for button in self.BUTTON_PINS.keys()}
-        hexpansion_states = {1: None, 2: None, 3: None, 4: None, 5: None, 6: None}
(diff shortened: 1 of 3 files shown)


emericklaw:
@MatthewWilkes as the person who created the button chord please can you cast your eye over this. My main concern is how I dealt with moving the LED update code into the event. I am not a python developer, maybe it should be somewhere else!

MatthewWilkes:
The light behaviour here is moving into a side effect of string coercion, rather than being a scheduled call, so it won't work (reliably).

It feels like these are independent features, to change the behaviour of the insertion event and to light the LEDs.

Do we know if the behaviour change is desirable? The current choice was just what was easy to do on-site, to be honest.

MatthewWilkes:
The latest firmware fixes hexpansion detection, so I think the best option is to remove the chord entirely. Thank you for putting the effort in to help this improve, though!

PR #111, opened 2024-06-05
Fix app store loading and error messages

emericklaw:
The pull request fixes two issues related to the app store.

The first commit removes a non-resolving URL that was causing the app store refreshing to take more than 15 seconds.

The second commit prevents the flashing of an "Unknown error" message when the index is loaded but the main menu is not created.


hughrawlinson (approved):

hughrawlinson:
Ugh I definitely had a short timeout on that first one at one point. Thanks!

PR #113, opened 2024-06-05
Improved keyboard

jake-walker:
When I got home and had to type in my WiFi credentials using the old keyboard, it took me so long I decided to crack out my laptop and run code to do it instead. This PR is a bit of an attempt to improve the speed of typing text in on the keyboard.

Each button on the badge corresponds to a group of letters, so the up button does letters **a** to **f**, once you press that, the buttons are reassigned letters through that range, so the up button now becomes **a**. The numbers are part of the uppercase letters (accessed with shift or caps) to keep the keyboard simpler. There's a 'meta' menu with backspace, shift, caps lock, symbols and cancel.

Screen recording of the improved keyboard

The keyboard is compatible with how the old one is used, so no other code needs to be changed.

As I wrote a function to label the buttons, I added yes/no labels to the `YesNoDialog` in the same way.

### DIFF
--- a/modules/app_components/dialog.py
+++ b/modules/app_components/dialog.py
@@ -4,7 +4,21 @@ import display
 from events.input import BUTTON_TYPES, ButtonDownEvent
 from system.eventbus import eventbus
 
-from .tokens import label_font_size, set_color
+from .tokens import label_font_size, set_color, small_font_size
+
+SPECIAL_KEY_META = "..."
+SPECIAL_KEY_DONE = "Done"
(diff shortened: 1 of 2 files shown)


jake-walker:
Just need to test this on my physical badge

jake-walker:
Can confirm the keyboard works on real hardware 🙂

npentrel:
I've just tested this as well - it works great! Thank you so much. I'll merge this now and it'll go into the next release (I don't know when that will be)

PR #114, opened 2024-06-07
Use production graphics for background image

kliment:
Now that the badges are out there, we can publish the production graphics. This changes the sim background image to the production one.


npentrel (approved):

npentrel:
Tested and working so will merge :D

PR #115, opened 2024-06-07
Mask wifi password in settings app

### LINKED ISSUE #105: Settings should hide wifi pw by default
Currently (1.6.0 of the firmware) the Settings page shows the wifi SSID and password, with both in plain text. It would be a good idea if the password had at least some extra level of obscurity (e.g. need to select that field with a button press in order to see the password).

jake-walker:
This PR masks the password shown in the settings app as well as on the keyboard when entering.

Fixes #105

### DIFF
--- a/modules/firmware_apps/settings_app.py
+++ b/modules/firmware_apps/settings_app.py
@@ -12,6 +12,13 @@ def string_formatter(value):
         return str(value)
 
 
+def masked_string_formatter(value):
+    if value is None:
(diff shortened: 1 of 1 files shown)


npentrel:
Can confirm this works on badge image

npentrel:
Thank you this is great! As with your other change, I'll merge this now and it'll go into the next release, whenever that is (I don't cut the releases)

Issue #81, opened 2024-05-31
Hexspansions with non eeprom I2C devices cause crash


ChrisDick:
Hexspansion has 2 I2C devices, one eeprom and one motor controller.

output log:

Registered async event handler for RequestForegroundPushEvent: _Scheduler - _handle_request_foreground_push
Registered async event handler for RequestForegroundPopEvent: _Scheduler - _handle_request_foreground_pop
Registered async event handler for RequestStartAppEvent: _Scheduler - _handle_start_app
Registered async event handler for RequestStopAppEvent: _Scheduler - _handle_stop_app
Registered async event handler for HexpansionInsertionEvent: HexpansionManagerApp - handle_hexpansion_insertion
Registered async event handler for HexpansionRemovalEvent: HexpansionManagerApp - handle_hexpansion_removal
Registered event handler for ButtonDownEvent: HexpansionManagerApp - handle_button_down
Registered event handler for ButtonUpEvent: HexpansionManagerApp - handle_button_up
Registered async event handler for PatternEnable: PatternDisplay - _enable
Registered async event handler for PatternDisable: PatternDisplay - _disable
Registered event handler for ButtonDownEvent: Launcher - _handle_buttondown
Registered async event handler for RequestStopAppEvent: Launcher - _handle_stop_app
Registered async event handler for InstallNotificationEvent: Launcher - _handle_refresh_notifications
Registered async event handler for ShowNotificationEvent: NotificationService - _handle_incoming_notification
Hexpansion inserted in port: 2
Found hexpansion header:
HexpansionHeader[
    manifest version: 2024,
    fs offset: 32,
    eeprom page size: 32,
    eeprom total size: 8192,
    vendor id: 0xCAFE,
    product id: 0xCAFF,
    unique id: 0,
    friendly name: Caffeine!,
]
Could not initialize eeprom: ('Non-contiguous chip addresses', [87, 90])
Hexpansion inserted in port: 3

walkerdanny:
I've done a bit of poking around and found what's going on here.

The helper script [`mount hexpansions.py`](https://github.com/emfcamp/badge-2024-software/blob/main/modules/scripts/mount_hexpansions.py) is calling:
https://github.com/emfcamp/badge-2024-software/blob/6900e39443cb653541ff73cc8d7156d0650755fa/modules/system/hexpansion/util.py#L57

This in turn constructs an EEPROM object, and in the constructor function for this object, it performs a scan for EEPROM devices:
[...]

PR #118, opened 2024-06-08
Improved hexpansion helper scripts

walkerdanny:
Changes to hexpansion helper scripts to work with EEPROMs that have the address 0x57 as well as 0x50 (previously 0x50 was hard coded). This now works with the Zetta ZD24C64A-XGMT EEPROM.

Change to `get_hexpansion_block_devices` to only look for one EEPROM on the bus, which I think fixes issue #81 where the script would throw an error when another I2C device was on the bus with a non-sequential address to the EEPROM.


walkerdanny:
The issue this fixes has been better troubleshooted in #81 and similar improvements made in #82, so I'm closing this.

Issue #89, opened 2024-06-01
Hexpansion detect is occasionally dodgy


thinkl33t:
Some badges have issues where the hexpansion detect isn't working consistently.

We *think* this is due to a software reset timing issue on the port expanders.

PR #122, opened 2024-06-09
Fix hexpansions not being consistently detected on hotplug

hedgehog1029:
For mysterious reasons the GPIO expanders seem to not always update the state of port 1 when doing a multi-byte read. If we instead do two single-byte reads, we appear to get the correct state 100% of the time.

The datasheet has some confusing lines about how you always need to send a write command before reading the port states - I believe `readfrom_mem` actually does this, but since it's not actually sending any data for the write, it's only technically writing to address 0, which only updates port 0 (hence why buttons work consistently - they're all port 0!)

### DIFF
--- a/modules/tildagonos.py
+++ b/modules/tildagonos.py
@@ -77,8 +77,10 @@ class _tildagonos:
 
     def read_egpios(self):
         for i in [0x58, 0x59, 0x5A]:
-            portstates = list(map(int, self.system_i2c.readfrom_mem(i, 0x00, 2)))
-            self.gpiodata[i] = tuple(portstates)
+            self.gpiodata[i] = (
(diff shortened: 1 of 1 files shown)


andypiper:
... fixes #89 ?

Robotmad:
I suspect this is also needed for set_egpio_pin.  Writing the the LS pins through the badge code doesn't work on some hexpansions - likely also where the pins are on port 1.  Whereas code writing only one byte (port) at a time is fine.

hedgehog1029:
Pushed an extra commit to fix `set_egpio_pin`, which was broken in other ways other than relying on the buggy multi-byte reads.

castaway:
Tested on a badge with dodgy hexpansion finding - works for me

PR #132, opened 2024-06-12
Fix menu width for wide items

npentrel:
This calculates the max font size for menu items so that menu items don't grow past the display width when in focus
Screenshot 2024-06-12 at 11 50 25

Screenshot 2024-06-12 at 12 37 51


MatthewWilkes:
Great idea.

PR #133, opened 2024-06-12
Ext gpio c

ChrisDick:
review required before merge. brings in low level c driver for port expanders and integrates in with existing APIs


modules/tildagon/pins.py
```diff
@@ -1,63 +1,65 @@
...
+    "5_LS_E": (0, 14),
+    "6_LS_A": (0, 7),
+    "6_LS_B": (0, 12),
+    "6_LS_C": (0, 13),
+    "6_LS_D": (0, 14),
+    "6_LS_E": (0, 15),
 }
 
 
 class Pin:
```
MatthewWilkes, line 37: Is the only reason to keep this wrapper to translate between the user-facing names and the internal names? If so, can we reduce this to just an `__init__`?

modules/tildagonos.py
```diff
@@ -37,73 +36,49 @@ def init_display(self):
...
-        self.system_i2c.writeto_mem(0x58, 0x7F, bytes([0x00]))
-        self.system_i2c.writeto_mem(0x59, 0x7F, bytes([0x00]))
-        self.system_i2c.writeto_mem(0x5A, 0x7F, bytes([0x00]))
-        # chip A - disable interrupts
-        self.system_i2c.writeto_mem(0x58, 0x06, bytes([0xFF, 0xFF]))
-        # chip A - set everything to input
-        self.system_i2c.writeto_mem(0x58, 0x04, bytes([0xFF, 0xFF]))
-        # chip A - switch mode to push-pull
-        self.system_i2c.writeto_mem(0x58, 0x11, bytes([0x10]))
+    def init_gpio(self): ...
```
MatthewWilkes, line 39: Is there anything left in this file that we want to use? I think we should add some deprecation warnings. Even if just the simulator.

modules/system/hexpansion/app.py
```diff
@@ -229,13 +229,10 @@ async def background_task(self):
...
         while True:
             with PerfTimer("indicate hexpansion insertion"):
-                tildagonos.read_egpios()
                 for i, n in enumerate(
                     [EPIN_ND_A, EPIN_ND_B, EPIN_ND_C, EPIN_ND_D, EPIN_ND_E, EPIN_ND_F]
                 ):
-                    hexpansion_present = not tildagonos.check_egpio_state(
-                        n, readgpios=False
-                    )
+                    hexpansion_present = not ePin(n).value()
```
MatthewWilkes, line 235: Is it worth bringing this out of the loop, so we don't re-instantiate them all every few milliseconds?

modules/tildagonos.py
ChrisDick: The only thing left is the LED power switch.

modules/system/hexpansion/app.py
ChrisDick: longer term I was planning on trying these as interrupt driven events to try and remove the i2c traffic.

thinkl33t:
please can this be rebased to allow pre-commit to pass?  Its *probably* not necessary given this doesn't touch the python much but i'd like to turn on enforcement of must-pass-ci :)

ChrisDick:
replaced with PR #166

PR #139, opened 2024-06-16
Reload the LED pattern generator on setting change

Jonty:
Without this it requires a reboop of the badge in order for the led pattern to actually change.

I have not tested this on an actual badge, only in the simulator - which doesn't have LED's!

This could be a lot more elegant if we allow apps to subscribe for callbacks when a specific settings key is updated.


MatthewWilkes:
Fixed issue when transitioning over the 'off' pattern, and LED display in simulator.

PR #157, opened 2024-06-17
Fix hexpansion insertion event text

MatthewWilkes:
The insertion events are not properly showing the hexpansion name. This confused people at camp, thinking the green border was part of the GCHQ.net app. This simplifies the code to use the port=0 code with the rotation applied, instead of trying to re-rotate


thinkl33t:
Have just tested this, appears to work well.

PR #159, opened 2024-06-17
Add support for a beta channel

MatthewWilkes:
# Description
This improves the OTA app to parse version numbers better, with the following rules:

- Major, Minor and Patch versions are parsed as numbers rather than strings, so v1.10.0 > v1.9.0
- Development versions where there's a 'number of commits ahead' item are counted as ahead of the version they're on
- If there's an rc after the patch number, it's treated as ahead of the previous version, so v1.9.0 < v1.10.0rc1 < v1.10.0rc1-1-aaaaaa < v1.10.0 < v1.10.0-1-aaaaaa

If there is an rc in the tag name, it won't overwrite the `latest` release, it will overwrite the `preview` release.

The setting `update_channel` will work as normal if unset or set to `latest`. Set it to `preview` to get release candidates.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -89,13 +89,13 @@ jobs:
           files: |
             micropython/ports/esp32/build-tildagon/micropython.bin
             micropython/ports/esp32/build-tildagon/tildagon.txt
-      - name: Create latest release for tags
+      - name: Create preview release for rc tags
         uses: "marvinpinto/action-automatic-releases@latest"
         if: ${{startsWith(github.event.ref, 'refs/tags/v')} && contains(github.event.ref, 'rc')}
         with:
           repo_token: "${{ secrets.GITHUB_TOKEN }}"
           automatic_release_tag: "preview"
-          title: "Latest release build"
+          title: "Latest preview build"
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -83,12 +83,22 @@ jobs:
...
           files: |
             micropython/ports/esp32/build-tildagon/micropython.bin
             micropython/ports/esp32/build-tildagon/tildagon.txt
+      - name: Create latest release for tags
+        uses: "marvinpinto/action-automatic-releases@latest"
+        if: ${{startsWith(github.event.ref, 'refs/tags/v')} && contains(github.event.ref, 'rc')}
+        with:
+          repo_token: "${{ secrets.GITHUB_TOKEN }}"
+          automatic_release_tag: "preview"
+          title: "Latest release build"
```
thinkl33t, line 98: Does this need to be "Latest preview build"

MatthewWilkes: Yes, fixed.

thinkl33t:
Doesn't crash and *should* work fine - lets try it and revert it if things explode :grin:

PR #161, opened 2024-06-19
Linting fixes

thinkl33t:
# Description

These changes should allow the linting and formatting stages to pass.


thinkl33t:
I've tested this on the simulator but not on a real badge.

Some things that need testing:

- fusb still works as expected / correctly
- [simple_tildagon guide](https://tildagon.badge.emfcamp.org/tildagon-apps/simple_tildagon/) still works correctly when followed

modules/fusb302b.py
```diff
@@ -462,9 +462,9 @@ def get_interrupts(self):
         Interruptb = self.i2c.readfrom_mem(
             self.ADDRESS, self.good_crc_sent_int.register, 1
         )[0]
-        Interrupt = self.i2c.readfrom_mem(
-            self.ADDRESS, self.bc_level_int.register, 1
-        )[0]
+        Interrupt = self.i2c.readfrom_mem(self.ADDRESS, self.bc_level_int.register, 1)[
+            0
+        ]
```
npentrel, line 467: I think this is the more generally accepted way of splitting this line
```suggestion
        Interrupt = self.i2c.readfrom_mem(
            self.ADDRESS, self.bc_level_int.register, 1)[0]
```

thinkl33t:
>     * fusb still works as expected / correctly
basic testing (making sure the cable plugged into itself still glows) seems to be fine

>     * [simple_tildagon guide](https://tildagon.badge.emfcamp.org/tildagon-apps/simple_tildagon/) still works correctly when followed
all code examples within still work

npentrel (approved):
LGTM % one nit

modules/fusb302b.py
thinkl33t: Agreed, i tried this (and the ones above are more readable than the autogenerated one here imo) but ruff keeps reformatting it back to its preferred option - are we happy enough to leave this as is and just merge accepting it looks a little odd?

npentrel: Funny. Sure feel free to merge

PR #166, opened 2024-06-22
egpio in c

ChrisDick:
port expander driver and integration, IN, OUT and PWM pin functions

# Description

replaces PR #133. Implements the c driver for the port expanders and integrates with c and python functions. ePin has similar API to machine.Pin with the addition of PWM open collector output. Initialised at boot and each hexpansion is reset to all inputs on removal.


MatthewWilkes:
I've read through this and tried it out a bunch. It's great.

I've removed some of the changes from #174, as the C implementation already set up the various modes in C, so the Python changes shadowed it.

I've also changed the print method of the ePin to make the repr match the init, and added one to the wrapper in tildagon.pin to do the same.

PR #167, opened 2024-06-22
Allow to install multiple apps

### LINKED ISSUE #120: Cannot install apps consecutively from the store
1. Open app store
2. Open StoreInstall
3. Install an app
4. Try to go back to StoreInstall (doesn't always work - can't press down sometimes)
5. Try to select another app (doesn't work - can't press down)

npentrel:
I think this fixes #120

### DIFF
--- a/modules/firmware_apps/app_store.py
+++ b/modules/firmware_apps/app_store.py
@@ -106,7 +106,10 @@ class AppStoreApp(app.App):
     def install_app(self, app):
         try:
             install_app(app)
-            self.update_state("main_menu")
+            if self.available_menu:
+                self.update_state("available_menu")
+            else:
+                self.update_state("main_menu")
             eventbus.emit(InstallNotificationEvent())
             eventbus.emit(ShowNotificationEvent("Installed the app!"))
         except MemoryError:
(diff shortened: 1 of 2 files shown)


modules/system/eventbus.py
```diff
@@ -40,10 +40,16 @@ def remove(self, event_type, event_handler, app):
         if app in self.handlers:
             if event_type in self.handlers[app]:
                 if event_handler in self.handlers[app][event_type]:
+                    print(
+                        "Removed event handler for {event_type}: {app.__class__.__name__} - {event_handler.__name__}"
```
npentrel, line 44: This still needs fixing before merge

ChrisDick:
tested this, works for me.

modules/firmware_apps/app_store.py
```diff
@@ -86,8 +92,13 @@ def background_update(self, delta):
                 return
             self.update_state("index_received")
         if self.to_install_app:
-            self.install_app(self.to_install_app)
-            self.to_install_app = None
+            # We wait one cycle after background_update is called to ensure the
```
npentrel, line 95: This feels a bit hacky but because update and draw may not be called while an app is installing I'm really not sure how else to fix it

npentrel: @MatthewWilkes why don't we merge this and then you can prettify it subsequently? It would be nice to have this functionality in

kliment (approved):

PR #168, opened 2024-06-22
Call cleanup before handlers for menu

npentrel:
With the way menu's work I don't think we should require users to call the `_cleanup method`. The underscore is supposed to mean it's internal anyway


npentrel:
@MatthewWilkes and @hughrawlinson how do we feel about this?

MatthewWilkes:
@npentrel So, I am not 100% on this, I think it's good for people not to have to call the cleanup method, but I think this would mean that all menu selects or cancels would have to tear down the menu, which I think limits the usages

PR #171, opened 2024-06-23
Allow app updating

### LINKED ISSUE #124: update mechanism for apps
an update mechanism for apps in general that isn't uninstall and reinstall.

npentrel:
Fixes #124.
Because we currently don't store the version (we do in the toml but adding regex checking into this feels error-prone), the first time a user runs upgrade they'll get all installed apps. The versions then get stored in the metadata.json and from there it will only list apps if there are apps to be updated.

This is easier to test if you rebase off of #167


npentrel:
@MatthewWilkes this is also ready to merge if you a ant to approve

kliment (approved):

ChrisDick (approved):

Issue #125, opened 2024-06-11
view descriptions of apps before installing (on badge)


npentrel:
I know you can view app descriptions on the web app store but would be nice if you can press right and get the description before installing

PR #172, opened 2024-06-23
Add information option to Menu widget

npentrel:
Adding the option to display information about a menu item when pressing right. Left or cancel will return you to the regular menu view. Adding it to the App store to allow users to read the description of apps before installing. We could do this from the app store instead but I think it is good functionality to live in the menu.

Addresses #125

Relies on #120


npentrel:
@MatthewWilkes do you want to review this as well? It's not really dependent on the other PRs just a bit more annoying to test and use

nic0lette (approved):
This is pretty cool!

The info shown (especially went scrolling it) looks a bit odd, but I think that's unrelated to this. (Text is written like the display is square, so the corners get chopped off =(

nic0lette:
I was thinking about this again, and I wonder if it wouldn't be possible (either in this PR or a follow-up) to add an "info" label to the "Right" button (when there's info that could be displayed) in order to help with discoverability?

MatthewWilkes (approved):
This seems to work well, and nice that it uses linearlayout for text scrolling

MatthewWilkes:
@nic0lette It's using the LinearLayout function to render the text, as that handles wrapping. It's intended for longer blocks of text (or widgets, as it's what runs the settings app also), but far from optimised. I think maybe starting it higher might help? A scroll-bar would also be nice, but it's hard to do it without sacrificing space. Displaying some text is a pretty common ask, so it wouldn't be a bad idea to improve it some. 

The code for this all is at https://github.com/emfcamp/badge-2024-software/blob/main/modules/app_components/layout.py if you're interested.

PR #173, opened 2024-06-23
Center Layouts menu items

npentrel:
Matthew - lmk what you think. We can also make this a configurable option rather than forcing everything centered?
layoutschange


MatthewWilkes:
I like it. Configurable might be nice, but I'm also fine with not offering options unless there's a reason users will need it.

MatthewWilkes (approved):

PR #174, opened 2024-06-23
Add init for ls pins to mimic behaviour of standard machine.pin

emericklaw:
# Description

This now better mimics the way standard machine.pins are initiated.
Also added checks when setting value to only apply to OUT pins.

### DIFF
--- a/modules/tildagon/pins.py
+++ b/modules/tildagon/pins.py
@@ -38,8 +38,13 @@ HEXPANSION_GPIOS = {
 
 class Pin:
     def __init__(self, pin_name, mode=-1):
+        self.IN = machine.Pin.IN
+        self.OUT = machine.Pin.OUT
         self.name = pin_name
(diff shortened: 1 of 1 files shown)


kliment:
Hello, and thank you for doing this!

Some related work is happening in https://github.com/emfcamp/badge-2024-software/pull/166 (and 133 before it, which 166 is a history-cleanup of) that should hopefully also make the external GPIO API-compatible. We would really like to move the gpio driver entirely out of Python so we can use it from the system libraries that are written in C, and once that is done the changes proposed here should, as far as we can see, no longer be necessary. Can you check whether this assumption is correct, and whether there's something missing from 166 that is present here? If not, the best way forward is probably to close this as a duplicate, but we wanted to get your input before doing so.

emericklaw:
Yes it looks like #166 should do the same.

PR #177, opened 2024-06-26
Improve async compatibility of app store

MatthewWilkes:
This moves some code from the background task to the main task, and refactors it to allow use of async methods. This looks like these tasks were in background from the idea that the background task runs 'in the background', rather than runs 'when the app is in the background'.

There are two calls that use requests.get, which block the update task. By refactoring slightly to use async, we can use async_helpers.unblock to move them to a thread.

### DIFF
--- a/modules/firmware_apps/app_store.py
+++ b/modules/firmware_apps/app_store.py
@@ -396,7 +396,8 @@ class AppStoreApp(app.App):
                 )
             except Exception:
                 self.update_state("no_index")
-            self.update_state("index_received")
+            else:
+                self.update_state("index_received")
 
             if self.to_install_app:
                 # We wait one cycle after background_update is called to ensure the
--- a/modules/firmware_apps/app_store.py
+++ b/modules/firmware_apps/app_store.py
@@ -4,10 +4,12 @@ import io
 import json
 import os
 import tarfile
+import time
 from tarfile import DIRTYPE, TarFile
 from typing import Any, Callable
 
 import app
+import async_helpers
 import wifi
 import shutil
 import machine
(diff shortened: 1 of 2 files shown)


modules/firmware_apps/app_store.py
```diff
@@ -255,6 +255,20 @@ def update(self, delta):
             self.prepare_available_menu()
         elif self.state == "installed_menu" and not self.installed_menu:
             self.prepare_installed_menu()
+        elif self.state == "refreshing_index":
+            try:
+                self.response = await async_helpers.unblock(
+                    get, render_update, APP_STORE_LISTING_URL
+                )
+            except Exception:
+                self.update_state("no_index")
```
npentrel, line 264: This would get changed in the next line wouldn't it

npentrel (commented):
I can't test this right now but I think there's one issue?

modules/firmware_apps/app_store.py
```diff
@@ -255,6 +255,20 @@ def update(self, delta):
...
+            self.update_state("index_received")
```
hughrawlinson, line 265: Yes, per Naomi's comment I think this ought to be like this?

```suggestion
            try:
                self.response = await async_helpers.unblock(
                    get, render_update, APP_STORE_LISTING_URL
                )
                self.update_state("index_received")
            except Exception:
                self.update_state("no_index")
```

modules/firmware_apps/app_store.py
MatthewWilkes: Fixed this with an else

MatthewWilkes:
Added a small bugfix to the launcher along with this, as it was causing trouble post install

hughrawlinson (approved):

PR #178, opened 2024-06-27
Better text wrapping

emericklaw:
# Description

Text was being wrapped in the middle of words.

Updated version only splits on spaces and if there is a word longer than will fit on one line then the word is hyphenated.

I would love to get the wrapping to work within a circular boundary but that will have to wait for another day!

## Normal Text - Before
Normal Text - Before

## Normal Text - After
Normal Text - After

## Long Word - Before
Long Word - Before

## Long Word - After
Long Word - After

## Settings - Before
Settings - Before

## Settings - After
Settings - After

### DIFF
--- a/modules/app_components/utils.py
+++ b/modules/app_components/utils.py
@@ -34,5 +34,4 @@ def wrap_text(ctx, text, font_size=None, width=None):
     for line in lines:
         lines = fill_line(ctx, line, font_size, width)
         wrapped_lines.extend(lines)
-    print(wrapped_lines)
     return wrapped_lines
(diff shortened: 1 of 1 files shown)


npentrel (commented):
The code looks good though I haven't tested it yet. I do think we should make hyphenation configurable. There may be cases in which adding a hyphen would be confusing. Displaying the WiFi ssid or password for example.

npentrel (approved):
Just tested, and will merge and address making the line wrapping character configurable separately

npentrel:
Thank you @emericklaw

Issue #83, opened 2024-05-31
Battery level incorrect


ChrisDick:
Battery Level incorrect. min and max voltage incorrect, Charge current could be used during charge to improve accuracy.

npentrel:
Is this still the case? My badge is no longer displaying above 100% charge levels...

ChrisDick:
yes, the power.BatteryLevel() is still returning higher than 100%. The battery meter app clamps it to 100%. Testing a fix now.

ChrisDick:
fixed in PR #179

PR #179, opened 2024-06-27
Improve battery level

ChrisDick:
# Description
fixes issue #83 
rework the battery level algorithm to be use different values for discharge and charge states and in the charge state use charge current profiles based on input current limits to calculate SOC. Reduce the charge termination current to 64mA to allow battery to charge to 4.2V.


kliment:
For battery lifetime I would suggest not charging up to 4.2. Do we have a reason to do so?

ChrisDick:
I've not actually changed the target voltage, just reduced the termination current. my comment is wrong, I misread issue #175 where the old max voltage used in the battery level api is 4.1V. While testing this change I saw 4.2V as the top of charge, so assumed the termination current change was allowing it to reach 4.2V. The PMIC defaults to 4.2V and we are not changing it, but we can if a lower target charge voltage is what we want.

npentrel (approved):
I can't say I understand the semantics here but the syntax looks fine... so LGTM?

ChrisDick:
I'm going to merge this in and if we want to change the target charge voltage, we can do it as a separate change.

PR #180, opened 2024-06-29
support self.pin during migration

ChrisDick:
re add the pin member with details for the I2C direct control of the port expanders to allow easier migration.

# Description

try to ease migration to the c gpio.


Robotmad (approved):
Thanks - looks like a very safe change.

npentrel (approved):

PR #185, opened 2024-07-16
enable pd comms

ChrisDick:
enable pd comms to add pd supply capabilities to power.SupplyCapabilities(), restrict the maximum voltage requested to 5V due to hardware limits and cap current request to max input current.


npentrel (approved):

ChrisDick:
I've altered the request to always use the first item in the list as this is guaranteed by the standard to be 5V with the highest current the supply can deliver at 5V.

kliment (approved):

PR #191, opened 2024-08-25
Egpio irq

ChrisDick:
# Description
Add interrupts to egpio, make buttons and hexpansion detection use interrupts, update sim to still use polling buttons and add irq() fake. disable IMU interrupts to allow development of these separately.


MatthewWilkes (approved):
I've done some extended playing with this now, and it seems to be working well and reliably. I think we merge, as the only things I can see that I would want to change are debt we've inherited from rushing to get things done in 24

Issue #109, opened 2024-06-05
Hexspansions/front face stay on after shutdown while charging


solexious:
I usually charge all my devices overnight while I sleep. But if you have the badge plugged into power when you shut it down it keeps the hexspansions powered and the front face on. This means my tent was lit up like an xmass tree, and I don't trust myself not to stand on it in the morning if I covered it with something to block the light.

Could the hexspansions and front face have their power cut after X seconds after shutdown is selected while connected to power for charging?

ChrisDick:
I've done a PR for turning off the LEDs and hexpansions when you power off the badge. However, powering off the badge disconnects the battery and prevents charging. A sleep charge mode is needed to do this, should be possible with an app too.

PR #193, opened 2024-09-11
turn off LEDs and hexpansions on power off

ChrisDick:
# Description

Turn off LEDs and Hexpansions when the badge is plugged in and powered off. requested in issue #109 but this doesn't allow charging, but does make the badge feel like it's been shutdown. An App might be better solution to the issue.


thinkl33t (approved):
LGTM

PR #194, opened 2024-10-22
Add app store category menu

jake-walker:
# Description

This splits the app store's listing of apps into categories. This is done based on the app store index and the categories are set by the app developers. Unfortunately, there's still a lot of them under "Badge", but it makes it slightly easier to navigate.

Screen Recording Demo


thinkl33t (approved):
LGTM

PR #197, opened 2025-02-28
web flasher: actively disable improv

thinkl33t:
We currently don't support IMPROV, but it keeps accidentally getting turned on when connecting to the badge over webserial.  This prevents the webserial terminal from being used, as it appends `IMPROV` to random characters

This MR should force it to be disabled.


Jonty:
Is there a reason we've not merged this?

PR #200, opened 2025-04-24
Update to Micropython 1.25, ESP-IDF 5.4.1

Molive-0:
# Description

Updates the entire repository to use the latest stable Micropython and ESP-IDF release.
Note that free flash space decreases from 12% to 9% with this update.

Also:
* Readme now recommends the build container be run with your own user rather than root (as per the ESP-IDF docs), which fixes some issues with build artifact permissions.
* Adds `scripts/hard-reset.sh`, which deletes all build artifacts and returns the submodules to how they were before `scripts/firstTime.sh` was run.
* Removes `CONFIG_FREERTOS_ENABLE_STATIC_TASK_CLEAN_UP=y` from the sdkconfig in order to fix the compile. Need to check if this is safe.

See [Jonty's comment](https://github.com/emfcamp/badge-2024-software/pull/200#issuecomment-2829028791) below for a quick summary of the Micropython changelogs.

## Todo

- [x] Check codebase for sys.exit() and raise SystemExit [(from Jonty's comment)](https://github.com/emfcamp/badge-2024-software/pull/200#issuecomment-2829028791). A quick Ctrl-f on the codebase seems to suggest that TildagonOS never does either of these things (aside from sys.exit()s in the sim), so we should be fine. Doesn't rule out any apps doing this, but I see no reason why they would.
- [x] Is it safe to remove `CONFIG_FREERTOS_ENABLE_STATIC_TASK_CLEAN_UP=y` from the sdkconfig, or is that integral to some part of TildagonOS? (see https://github.com/emfcamp/badge-2024-software/pull/200#discussion_r2143131041) Config option was added in [this commit](https://github.com/emfcamp/badge-2024-software/commit/f78c9307cc7ee1fa2e96eb964e9ace00a40809ca) by Matthew Wilkes, but documentation suggests it's only for if you have a `vPortCleanUpTCB` function, and I don't think we do.


Jonty:
Quickly skimming the .24 and .25 release notes for things we need to check or should take advantage of:

## [v1.24.0](https://github.com/micropython/micropython/tree/v1.24.0)
### To check
* The behaviour of sys.exit() and raise SystemExit (which are equivalent) has been changed: previously this would terminate the running script and drop to the REPL, but now it terminates the running script and triggers a soft reset of the device. This is more consistent with the unix port, which exits completely on a SystemExit exception. This change in behaviour is a breaking change and may impact certain applications; see commit https://github.com/micropython/micropython/commit/69c25ea8653566ec97690b5121bd10b753c89426 for details.
### Interesting
* In the networking sub-system, a new API for configuring IP addresses and related settings has been added: the global network.ipconfig() function, and a nic.ipconfig() method on individual network interfaces. This new API supports both IPv4 and IPv6, with much more control compared to the original nic.ifconfig() method; the latter is still available for backwards compatibility, but ipconfig() is now preferred moving forward.

## [v1.25.0](https://github.com/micropython/micropython/tree/v1.25.0)
### Interesting
* After more than three years in development, the "ROMFS" feature has been finalised, its filesystem format specified, and the VFS driver and supporting code are included in this release of MicroPython. This feature builds on bytecode version 6 (available for many years now), which supports executing bytecode in-place, that is, without the need to copy it to RAM. ROMFS defines a read-only, memory-mappable, extensible filesystem that can contain arbitrary resources, including precompiled mpy files, and allows executing bytecode directly from the filesystem. This makes importing significantly faster and use a lot less memory. Also, data resources such as fonts can be used in-place on a ROMFS without loading into RAM.
* The MicroPython native linker mpy_ld.py now includes support for linking in static libraries automatically. This allows the native-module build scripts to look for required symbols from libraries such as libgcc and libm that are provided by the compiler. This now makes it possible to use standard C functions like exp() in native modules. Also, native modules now support 32-bit RISC-V code.
* Dynamic USB device support is now enabled on ESP32-S2 and ESP32-S3 MCUs, allowing configuration of the USB device at runtime. I2S has been enabled on all ESP32-C3 boards, the Pin.toggle() method has been added, and the I2C bus identifier is now an optional argument (by default, bus 0 is used). Additionally, memory management has been improved for the allocation of TLS sockets to attempt to automatically free any existing unused TLS memory when needed.

Molive-0:
Annoyingly, the CI has failed to compile this. I'll look into it tomorrow.

Molive-0:
This branch should now build in CI, as a test I ran on my laptop locally seems to work. It did require removing the tildagon_usb driver, however it seems that it never did anything in the first place (let alone anything to do with USB)

ChrisDick:
There's been a change to the idf where freertos.h no longer includes idf_additions.h. I think #include "freertos/idf_additions.h" should be added instead of removing the usb from the build as the dynamic usb device support that this change brings in may fix the issues with the usb branch. 

Also the docker pull isn't working, probably needs @MatthewWilkes to create it.

tildagon/sdkconfig.board
```diff
@@ -26,5 +26,3 @@ CONFIG_BOOTLOADER_APP_ROLLBACK_ENABLE=y
 CONFIG_APP_EXCLUDE_PROJECT_VER_VAR=n
 CONFIG_APP_EXCLUDE_PROJECT_NAME_VAR=n
 CONFIG_APP_PROJECT_VER_FROM_CONFIG=y
-
-CONFIG_FREERTOS_ENABLE_STATIC_TASK_CLEAN_UP=y
```
Molive-0, line 30: I am personally unsure on the ramifications of this change to the sdkconfig. It was required to make the compile work, and the firmware appears to behave, but I can't help but feel that this is needed somewhere.
All changes made as part of this review were to fix existing features during the upgrade rather than remove them, but no matter what I did this did not seem to work.

ChrisDick:
Since this is now conflicted( I can't resolve on your branch), can it be updated to Micropython v1.26 and ESP-IDF v5.4.2?

Molive-0:
It appears that the 1.26 update doesn't work immediately, and will take time to debug the issues around it.
To this end I suggest that I return this PR back to 1.25, and then create a separate PR with updates to 1.26.

ChrisDick:
yeah, good suggestion, thanks for trying it.

tildagon/sdkconfig.board
ChrisDick: there's a warning when building with previous: #warning "CONFIG_FREERTOS_ENABLE_STATIC_TASK_CLEAN_UP is deprecated. Use CONFIG_FREERTOS_TASK_PRE_DELETION_HOOK instead." for mp1.25 the pre deletion hook is set to y in sdkconfig.base. I think it's right to remove it.

Molive-0:
> Just skimmed through this and I'm unclear on the current state - does this actually work? Can we merge and upgrade to 1.25?

As far as I'm aware it still works. I'll check it again after work

Molive-0:
todos are now marked complete

PR #201, opened 2025-04-24
SDK Config Tweaks

Molive-0:
# Description

Tweaks to the sdkconfig.board file which increases the free flash space from 6% to about 14%.

Also increases cache size.


MatthewWilkes (approved):
Nice, thanks!

PR #204, opened 2025-05-17
Update power supply capability functions to use float for voltage values

emericklaw:
# Description
Update power supply capability functions to use float for voltage values.

When using the voltage values from `power.SupplyCapabilities()` they were being displayed as int instead of floats, 4.90V would be displayed as 4.00V.

With this change it correctly displays as 4.90V

### Example code

```python
import power

SupplyCapabilities = power.SupplyCapabilities()[0]
ctx.move_to(0, 0).text("Voltage: {:.2f}".format(SupplyCapabilities[2]) + "V")
```

## Before

image

## After

image


ChrisDick:
Can you change the divides to 1000.0F, otherwise the integer rounding will still be present on the battery and variable supply cases. 
Also the minimum battery voltage on line 171 needs to be a float too.

Issue #202, opened 2025-05-17
Cannot update firmware using OTA


emericklaw:
**Describe the bug**
When updating the firmware the error `ESP_ERR_HTTP_CONNECT` is shown.

Digging into this I see this:

```
<function head at 0x3c2b16e0> ('https://github.com/emfcamp/badge-2024-software/releases/download/latest/micropython.bin',) {'allow_redirects': False}
Periodic
...
Periodic
<Response object at 3c248920>
<Response object at 3c248920>
Starting thread
Periodic
In thread
<function> (<bound_method>, 'https://objects.githubusercontent.com/github-production-release-asset-2e65be/751434638/a0f4124c-f2df-4383-9b73-bb693e9fd1e6?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=relea{}
Periodic
Exception: (-28674, 'ESP_ERR_HTTP_CONNECT')
(-28674, 'ESP_ERR_HTTP_CONNECT')
(-28674, 'ESP_ERR_HTTP_CONNECT')
```

I thought this might be related to certificate issues, however the certificate for `githubusercontent.com` has USERTrust RSA Certification Authority as its root which, if I am looking in the right place is in the list https://github.com/espressif/esp-idf/blob/v5.2.1/components/mbedtls/esp_crt_bundle/cacrt_all.pem

I altered the url here to point to a copy of the .bin on a personal website and it flashed fine so there is something weird going on with the GitHub URL.

https://github.com/emfcamp/badge-2024-software/blob/f5a3c5824acf18a229ad051996029b107b7da95f/modules/system/ota/ota.py#L153

Version: 1.9.0

**To Reproduce**

1. Go to 'Update'
2. Wait for update process to start
3. Error `Failed: (-28674, 'ESP_ERR_HTTP_CONNECT')` is shown

**Expected behavior**
The firmware OTA completed successfully.

dpslwk:
I believe `CONFIG_MBEDTLS_CERTIFICATE_BUNDLE_DEFAULT_CMN` is set, so only certs from https://github.com/espressif/esp-idf/blob/v5.2.1/components/mbedtls/esp_crt_bundle/cmn_crt_authorities.csv are include from the full `cacrt_all.pem` you linked to

not sure if that still include the needed cert for the url to be built into the firmware

dpslwk:
https://github.com/espressif/esp-idf/issues/3218
[...]

PR #205, opened 2025-05-18
Use full root certificate bundle to fix broken OTA updates

emericklaw:
# Description

The OTA update is broken as discussed in #202 which relates to certificate issues with `objects.githubusercontent.com`.

The current build only include the common root certificates using `CONFIG_MBEDTLS_CERTIFICATE_BUNDLE_DEFAULT_CMN=y`
Changing this to `CONFIG_MBEDTLS_CERTIFICATE_BUNDLE_DEFAULT_FULL=y` gets OTA updates working again.

This does increase the firmware size by 46k but takes the root certificate coverage from 90% to ~100%.

The other alternative is to include the missing root certificate for `Comodo AAA Services root` which has been done in the 5.4.1 version of ESP-IDF.

https://github.com/espressif/esp-idf/commit/a83a0ab02be248cd45fbc80131db274d9b883ba0

Either via a patch or upgrading ESP-IDF to 5.4.1 which is covered in #200 

This unfortnately means that all current badge users will have to reflash using the web flasher.


MatthewWilkes:
This works well, thanks!

MatthewWilkes (approved):

PR #206, opened 2025-05-19
Tries OTA via HTTPS then flips to HTTP on failure then repeats

### LINKED ISSUE #202: Cannot update firmware using OTA

emericklaw:
# Description

This PR changes the OTA download to first try HTTPS then on failure tries HTTP and repeats flipping between HTTPS/HTTP.

Fixes #202

### DIFF
--- a/modules/system/ota/ota.py
+++ b/modules/system/ota/ota.py
@@ -135,7 +135,7 @@ class OtaUpdate(App):
         # window.println()
         # line = window.get_next_line()
         self.confirmed = False
-
+        self.tryHttps = True
         retry = True
         self.status.value = "Searching for OTA"
 
@@ -151,6 +151,9 @@ class OtaUpdate(App):
             )
             response = await self.task
             url = response.headers["Location"]
+
+            if not self.tryHttps:
+                url = url.replace("https://", "http://")
 
             """
             self.task = async_helpers.unblock(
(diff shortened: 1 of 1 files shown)


thinkl33t (approved):
This solves an issue we've had a few times in the past - github changing their TLS implementation in a way that breaks our tiny TLS implementation.

While I'm not normally a fan of fail-to-insecure i don't see any other options in this case.

PR #213, opened 2025-08-07
update accelerometer

ChrisDick:
Add temperature_read() and access to the device on i2c, readfrom() and writeto() to allow the system i2c devices to be secured and still have access to the imu. Add detection for IMU device type and implement the LSM6DS3 device from the prototype badge with an id() to identify which device is on board. any alternate imu must support accel, gyro, temperature and step functions.

# Description

Please include a summary of the change and what is does. Please also include relevant motivation and context.
add support for the prototype badge accelerometer to let us use a cheaper device if the BMI270 isn't sponsored.


MatthewWilkes (approved):
Tested and working

PR #224, opened 2025-08-26
Add toggle for update channel

MatthewWilkes:
# Description

Make it easier for users to select the preview update channel to test OTAs, without needing them to use mpremote and write Python


nic0lette (approved):
Nice!  =D

thinkl33t (approved):
LGTM

PR #210, opened 2025-06-10
Custom background

ChrisDick:
# Description

feature to allow customised backgrounds, 2 are in the badge firmware and more can be downloaded from the app store. a background app uses the draw and update functions called from firmware apps and accessible to user apps. user backgrounds are stored in a backgrounds folder. 

https://github.com/user-attachments/assets/f8fb7359-8ae1-4786-98dc-b6b27c0469a3

PR #228, opened 2025-08-27
fix issue 221

ChrisDick:
# Description

I found that some apps don't always enable the pattern before minimising, Starfield for example, so I made minimise do it to make the main menu as it was before starting the app.

### DIFF
--- a/modules/app.py
+++ b/modules/app.py
@@ -3,6 +3,7 @@ import time
 
 from system.eventbus import eventbus
 from system.scheduler.events import RequestForegroundPopEvent
+from system.patterndisplay.events import PatternEnable
 
 
 class App:
@@ -51,3 +52,4 @@ class App:
 
     def minimise(self):
         eventbus.emit(RequestForegroundPopEvent(self))
+        eventbus.emit(PatternEnable())
--- a/modules/app.py
+++ b/modules/app.py
@@ -3,7 +3,6 @@ import time
 
 from system.eventbus import eventbus
 from system.scheduler.events import RequestForegroundPopEvent
-from system.patterndisplay.events import PatternEnable
 
 
 class App:
@@ -52,4 +51,3 @@ class App:
 
     def minimise(self):
         eventbus.emit(RequestForegroundPopEvent(self))
-        eventbus.emit(PatternEnable())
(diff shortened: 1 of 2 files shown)


emericklaw:
Would this possibly conflict with custom background apps that other own LED effects?

Maybe a way that an app can flag that it is controlling the LEDs.

emericklaw:
Would this possibly conflict with custom background apps that other own LED effects?

Maybe a way that an app can flag that it is controlling the LEDs.

nic0lette:
> I keep going back and forth between wondering if we should integrate them better into the core OS or just get out of their way

Now that we have "backgrounds" from #210 I wonder if it's reasonable to say that apps that want to control the LEDs while they aren't active should either a) be a background or b) actively do something* to keep control of them when the app is minimized

The problem I see with "get out of their way" is that this means that all the apps that change LEDs inside the app (AKA the OTA updater and other similar things) have to re-enable the user's pattern on exit, which is (I would guess) more work for the more common use case? (Control LEDs in the foreground only)

\* either a setting or maybe extend "App" to have a protected method that is called when minimizing that, by default, restores the user's pattern, so apps can just `pass` to skip it?

Jonty:
I think we need to build this into the core really.

Perhaps something like:
* Foreground app should always have control of the LEDs if it requests it (but can release it, if it just uses them for a short time for progress bars or something)
* If the foreground app has not requested it, a background app can update them
* If multiple background apps have registered for control of the LEDs, control goes to the background app that was last interacted with/started

This would make most sense to the user, as it's whatever they are doing/last did that keeps updating them.

Perhaps something like `requestLEDs()` to attempt to get control of them, and `releaseLEDs()` to let someone else take over. If you've requested them and another app has taken control we could either have LED writes fail, or just have them ignored?

ChrisDick:
so the stats from the 75 available apps are:
31 apps disable LEDs
11 apps enable LEDs

of the 31, 2 apps use LEDs in background tasks:
PrideCircles, lets their led sequence continue to run when minimised.
emf cave, shows a game is still waiting for a response to play again.

so yes, it does affect some apps, but both of these would be ok with being able to change the current pattern being displayed.

Since we have PatternDisable and PatternEnable events, all we need is the API in PatternDisplay for apps to change the background pattern without changing the user setting. This will give the last app wins approach. I'm working on a user pattern that can be selected in settings, similar to the background feature, so I'll include the API for apps to use. 

I'll change this PR to only address issue 221 and do another PR for this, once we got the two apps to use the new API. We'll need to release it before they are updated and this makes it a good candidate for 2.0 with the MP breaking change if we can do it in time.

MatthewWilkes (approved):

Issue #234, opened 2025-08-31
LED pattern doesn't restart after app minimise


ChrisDick:
**Describe the bug**
2/3rds of apps don't re-enable the LED pattern on minimise causing the LEDs to look like they've stopped working.

Version: v1.9.1

**To Reproduce**
start an app that disables pattern but does not re-enable it.

**Expected behavior**
LEDs keep blinking.

**Additional context**
discussed in #228. Currently there is no way of users controlling LEDs when not in their own app, so the background task is used. I proposed adding a custom pattern feature and an api to change which pattern is being displayed without changing the user settings so that apps can control LEDs in the background.

PR #235, opened 2025-08-31
custom LED patterns

ChrisDick:
Feature to allow user LED pattern apps. also add API for normal apps to set the LED pattern without changing the settings

# Description

Please include a summary of the change and what is does. Please also include relevant motivation and context.
see #228 and #234

### DIFF
--- a/modules/firmware_apps/app_store.py
+++ b/modules/firmware_apps/app_store.py
@@ -337,7 +337,6 @@ class AppStoreApp(app.App):
         else:
             selected_app = selected_app[0]
         selected_app_module = selected_app["path"]
-        print(selected_app_module)
         selected_app_fs_path = "/" + "/".join(selected_app_module.split(".")[0:-1])
         print(f"Selected app fs path: {selected_app_fs_path}")
         shutil.rmtree(selected_app_fs_path)
--- a/modules/firmware_apps/app_store.py
+++ b/modules/firmware_apps/app_store.py
@@ -20,25 +20,17 @@ from requests import get
 from system.eventbus import eventbus
 from system.launcher.app import (
     APP_DIR,
-    list_user_apps,
     load_info,
     InstallNotificationEvent,
 )
 from system.notification.events import ShowNotificationEvent
 from app_components.background import Background as bg
-from firmware_apps.settings_app import BG_DIR
+from firmware_apps.settings_app import BG_DIR, PAT_DIR
 
 
 def dir_exists(filename):
     try:
         return (os.stat(filename)[0] & 0x4000) != 0
-    except OSError:
-        return False
-
-
-def file_exists(filename):
-    try:
-        return (os.stat(filename)[0] & 0x4000) == 0
     except OSError:
         return False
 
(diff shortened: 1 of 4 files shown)


modules/firmware_apps/settings_app.py
```diff
@@ -11,6 +11,7 @@
 from system.scheduler.events import RequestForegroundPushEvent
 
 BG_DIR = "/backgrounds"
+PAT_DIR = "/pattern"
```
benclifford: for consistency, this could be plural /patterns like /backgrounds and /apps

modules/system/patterndisplay/app.py
```diff
@@ -2,29 +2,78 @@
...
+            path = f"/pattern/{self.pattern[1]}/app.py"
+            print(path)
+            try:
+                os.mkdir(PAT_DIR)
+            except OSError:
+                pass
+            if path_isfile(path):
+                try:
+                    _patternpath = f"pattern.{self.pattern[1]}.app"
+                    _patternclass = "__Pattern_export__"
```
benclifford: for consistency with __app_export__ this could be lowercase

benclifford:
I just tried this out because I was hitting the problems with too many of my own apps wanting to be in charge of the LEDs. It merges cleanly into `main` right now. It was pretty straightforward to port one of my apps to be a pattern, and I like both the UI and the amount of `App` code I could delete.

:+1:

modules/system/patterndisplay/app.py
```diff
@@ -2,29 +2,78 @@
...
+                    _pmodule = __import__(
+                        _patternpath,
+                        globals(),
+                        locals(),
+                        [
+                            _patternclass,
+                        ],
+                    )
+                    _pclass = getattr(_pmodule, _patternclass)
+                    self._p = _pclass()
```
benclifford: If this raises an error due to my terrible user-provided pattern implementation having a bug in its `__init__`, then when I power up my tildagon, I get a blank screen/dead tildagon that needs external intervention (to stop this pattern loading at startup) - it crashes `main.py`.

For example my pattern that expects wifi to be up:

```
  File "main.py", line 24, in <module>
  File "system/patterndisplay/app.py", line 27, in __init__
  File "system/patterndisplay/app.py", line 56, in load_pattern
  File "pattern/wiflame/app.py", line 18, in __init__
OSError: Wifi Internal Error
```

I'm happy to do some work here -- it's definitely going to bug me again while I play with patterns.

modules/system/patterndisplay/app.py
ChrisDick: I've updated this. I'll still need to update my app and test it.

modules/firmware_apps/app_store.py
```diff
@@ -52,35 +44,43 @@ def file_exists(filename):
...
             app.update(metadata)
             apps.append(app)
         return apps
 
 
+def list_all_apps():
+    return (
+        list_apps(APP_DIR, "__app_export__")
+        + list_apps(BG_DIR, "__Background__")
+        + list_apps(PAT_DIR, "__Pattern_Export__")
```
MatthewWilkes, line 80: I think that technically this should be `__pattern_export__`

modules/system/patterndisplay/app.py
```diff
@@ -2,29 +2,83 @@
...
+            path = f"/pattern/{self.pattern[1]}/app.py"
+            print(path)
+            try:
+                os.mkdir(PAT_DIR)
+            except OSError:
+                pass
+            if path_isfile(path):
+                try:
+                    _patternpath = f"pattern.{self.pattern[1]}.app"
+                    _patternclass = "__pattern_export__"
```
MatthewWilkes, line 45: ... however, this bit hard-codes the callable, so we don't need to worry about the mismatch above.

MatthewWilkes (approved):
There's a little confusion about capitalisation, but it doesn't stop it working. I've tested with a custom pattern, and it enumerates and selects it correctly.

PR #236, opened 2025-08-31
Improve settings app performance

MatthewWilkes:
# Description

Improve reliability of the settings app by:

1. Remove extraneous draw call
2. Use micropython optimisation decorator on word wrap util function
3. Avoid calling draw() method for items that will be out of the clip region


MatthewWilkes:
Replaced by #259

PR #237, opened 2025-09-17
add ePin toggle()

ChrisDick:
adding toggle() as machine.Pin has had a toggle function added in v1.25.


MatthewWilkes (approved):
Tested and working

PR #241, opened 2025-12-06
use docker build container in ci

hughrawlinson:


hughrawlinson:
This appears to work, but I haven't verified the actual build output,. If someone can verify that [the actual output](https://github.com/emfcamp/badge-2024-software/actions/runs/19996028555/artifacts/4787542363) is correct and valid, then I guess we should merge this as it speeds up the build quite a bit - I guess that's part of the benefit of the build container.

ChrisDick (approved):
I've run this on a badge and it looks like it's fine. I'm not able to rebuild locally and get the same binary, but I don't think this is because of this change and this is a good step to getting consistent builds.

thinkl33t (approved):

PR #244, opened 2026-01-13
Add Emote events

benclifford:
# Description

These are intended for foreground apps to signal to other components (for example hexpansion drivers) that something interesting, good or bad, happened in the user interface, so that the other component can present that in its own way to the user.

This PR has three commits:

* base event definitions in events.emote
* modifications to app store and sponsors to make some emotes
* an example emote driver that flashes the back LEDs when an emote happens.

The app and driver commits are more for discussion about what any default emote behaviour should be rather than immediate merge.

I will also open a separate PR with only the event definitions, intended for actual merge.
        
Guidelines for use:
    
Only send emotes when you are the foreground app: these events are intended to be UI output like writing to the screen is.
    
Send at most a handful of emotes per minute. These events might generate relatively long behaviours on the order of
a few seconds, for example playing a short jingle on an audio hexpansion.

### DIFF
--- a/modules/app_components/background.py
+++ b/modules/app_components/background.py
@@ -1,4 +1,5 @@
 import settings
+from events.emote import EmoteNegativeEvent
 from system.backgrounds.emflogo import EmfLogo
 from system.backgrounds.wigglinghexagons import WigglingHexagons
 from system.eventbus import eventbus
@@ -38,6 +39,7 @@ class _Background:
                             message=f"Background {self.selection[0]} has crashed"
                         )
                     )
+                    eventbus.emit(EmoteNegativeEvent())
                     self.runner = None
             else:
                 print("path not a file: " + path)
@@ -58,6 +60,7 @@ class _Background:
                         message=f"Background {self.selection[0]} has crashed"
                     )
                 )
+                eventbus.emit(EmoteNegativeEvent())
                 self.runner = None
             ctx.restore()
         else:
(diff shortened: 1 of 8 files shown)


thinkl33t:
Merged @ #245

thinkl33t:
Misread #245 - reopening

benclifford:
I'm still working on this - just took a few months off to go travelling etc

MatthewWilkes:
@benclifford Was this something you were still keen on getting in for this year?

benclifford:
I was planning on that originally. but I realise suddenly we are very close to the event. I'll bring it all up to date in the next day or so and see where it is at.

benclifford:
I've brough this up to where I think I am confortable with whats achievable right now. The commits in the branch right now do this in sequence:

i) Move the handling of the back-of-board LEDs into a manager app, separate from the hexpansion code, similar-but-different to the pattern manager. (no functionality change intended here).  knowledge of hexpansion insertion comes via the existing eventbus HexpansionInserted/Removed events.

ii) Make the backleds manager also understand emote events, and perform a red/green flash on posive/negative.

iii) Add a bunch of emote events to onboard software. Mostly app crashing negative events, with a positive event on successful app install.

iv) Add a settings option next to "mirror patterns" to turn emote-backleds off. I set it on by default because I love excessive LED flashing / to encourage people to realise emote events are there.

There is scope for more functionality/configuration of the back leds manager (e.g. more principled other-app takeover, or permanent mirroring for people who have no hexpansions) - that PR lays the groundwork for that but is for the future, not this PR.

ChrisDick (approved):

PR #245, opened 2026-01-16
Add emote event definitions

benclifford:
These are intended for foreground apps to signal to other components (for example hexpansion drivers) that something interesting, good or bad, happened in the user interface, so that the other component can present that in its own way to the user.

Guidelines for use:

Only send emotes when you are the foreground app: these events are intended to be UI output like writing to the screen is.

Send at most a handful of emotes per minute. These events might generate relatively long behaviours on the order of a few seconds, for example playing a short jingle on an audio hexpansion.

### DIFF
--- a/modules/events/emote.py
+++ b/modules/events/emote.py
@@ -1,0 +1,10 @@
+from events import Event
+
+
+class EmoteEvent(Event): ...
+
+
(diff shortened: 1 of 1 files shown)


sodoku:
I think it would be good to also add this to the docs.

PR #248, opened 2026-01-18
add pd Vendor messages and badge detection

ChrisDick:
# Description

Cleaned up the stack and added host and device vendor messages to pd. This allows users to send and receive vendor messages on both ports once USB attach is complete for that port. 

If a device is attached to a host port a vendor message is sent. If message is received by the badge on the usb in, it detects the other badge and connects as a device, it responds with the same message to let the other badge know it is connected as a host.

includes a helper to construct a vendor header and events to signal a badge has been connected and messages received.

It also contains code for prime and double prime messages, but there is an issue with the packet creation and the end of packet symbol is being inserted by the fusb302 before it stuffs the CRC in the packet. hopefully this will be enabled in a future PR.


ChrisDick:
updated the api to the following, using separate classes as host may be extended with prime messages, which a device deosn't send:

```
>>> from pd import Device, Host
>>> dir(Device)
['__class__', '__name__', '__bases__', '__dict__', 'badge_connected', 'connected', 'get_vendor_msg', 'pd_enabled', 'send_vendor_msg']
>>> dir(Host)
['__class__', '__name__', '__bases__', '__dict__', 'badge_connected', 'connected', 'get_vendor_msg', 'pd_enabled', 'send_vendor_msg']
```

thinkl33t (commented):
Seems sensible enough though my C++ skills have definitely atrophied a bit 🙃

Couple of queries, may be nothings.

MatthewWilkes:
@ChrisDick I think there's a call to `host_send_badge_id` missing on pd connect on the Host port. If I manually call this, then the detection and messages work perfectly, but without it the badges don't detect each other.

ChrisDick:
line 244 in tildagon_power.c has the host sending the message on the attach event. This then triggers the device to send it in response. the pd_handler doesn't send the host message on connect. do you get a connect notification when you plug it in?

MatthewWilkes:
I do not, only when I manually send the message from the host. I also don't get the badge_connected returning true or the mp event. I get the same with either of my badges as host.

Hmm, I wonder what I could try to debug.

MatthewWilkes (approved):
@ChrisDick On the basis that this is working for you, let's merge it and get going on `v2.0.0-a1`. The fact it's not working for me without the explicit call to send the message makes me think there is some flakiness somewhere, but let's not faff around with polishing it any more if we can get it into alpha testers hands? We can always fix the flakiness in another PR

ChrisDick:
I've moved the host sending badge id to the hot attach handler in python. This is giving it some more time to start pd on the device and respond. There is still an issue with detaching, but I think this is outside of the PD comms so shouldn't stop this going in.

MatthewWilkes:
@ChrisDick I would merge this, but it's conflicting for me

PR #261, opened 2026-04-06
toggle repl type

ChrisDick:
# Description
default to the original repl and add a toggle between asyncio version.

ideally needs to go into rc before release.

### DIFF
--- a/modules/firmware_apps/settings_app.py
+++ b/modules/firmware_apps/settings_app.py
@@ -62,6 +62,7 @@ BRIGHTNESSES = [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0]
 CHANNELS = ["latest", "preview"]
 REPL_TYPES = ["Default", "asyncio"]
 
+
 class SettingsApp(app.App):
     def __init__(self):
         self.layout = layout.LinearLayout(items=[layout.DefinitionDisplay("", "")])
@@ -265,14 +266,13 @@ class SettingsApp(app.App):
                                 idx = 0
                             print(f"{repl_type} {idx}")
                             settings.set("repl_type", REPL_TYPES[idx])
-                            
+
                             await self.update_values()
                             await render_update()
                             return True
                         return False
-                    entry = layout.ButtonDisplay(
-                        "Toggle", button_handler=_repl_toggle
-                    )
+
+                    entry = layout.ButtonDisplay("Toggle", button_handler=_repl_toggle)
                     self.layout.items.append(entry)
 
             async def _button_event_w(event):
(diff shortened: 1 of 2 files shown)


benclifford:
I merged this into my firmware fork. These are the new behaviours I expected and how this new firmware did not behave as I expected:

i) by default, no repl appears when you use `minicom` to get a console.
`mpremote fs ls` crashes the badge firmware and lets you do filesystem access.
After doing that, `minicom` gives you the traditional repl.

The is the behaviour from before my PR #254 which added the aiorepl.

This i) behaviour is consistent with what I expected.

ii)

Choose Settings / REPL type toggle to set type to `asyncio`.

Without doing a reboot, the badge behaves as (i) above:

Now reboot the badge using soft reboot (exit the default repl above)

As this causes the firmware to restart, aiorepl appears with minicom.

At this point, I expected `mpremote fs ls` to hang - because aiorepl is somehow in charge of the console.

*however* actually `mpremote fs ls` did not hang and instead did return an `ls`. At the same time... the aiorepl still presents itself and can be used, and the badge firmware continues to run.

This is exactly the behaviour I would have liked but I wasn't seeing earlier (and the motivation for this PR #261 in the first place).

So I am a bit confused.

ChrisDick:
ok, so it doesn't change the way mpremote works with asyncio repl, but does let thonny(etc?) connect and mount the filesystem with default settings.

benclifford:
without digging into the protocol, I have a feeling that random apps and firmware outputting stuff using `print` is going to screw up some `mpremote` stuff and that having the aiorepl turned off in that situation is the right thing  - and maybe what happened to break it for me earlier on when it didn't work for me.

ChrisDick:
I have a working app, can you give it a try to see if it does as expected:
```python

import asyncio 
import aiorepl
from app import App
from app_components import clear_background
from events.input import BUTTON_TYPES, ButtonDownEvent


class ReplApp(App):
    def __init__(self, config=None):
        self.started = False
        eventbus.on_async(ButtonDownEvent, self._button_handler, self)

    async def _button_handler(self, event):
            self.minimise()
        elif BUTTON_TYPES["CONFIRM"] in event.button:
            self.started = True
        return True

    async def run(self, render_update):
        while True:
            await render_update()
            if self.started is True:
                await render_update()
                repl_task = asyncio.create_task(aiorepl.task())
                await repl_task

    def draw(self, ctx):
        clear_background(ctx)
        if self.started:
            ctx.rgb(1, 0, 0).move_to(-80, 0).text("Enabled")
        else:
            ctx.rgb(1, 0, 0).move_to(-80, 0).text("Enable asyncio REPL")

__app_export__ = ReplApp


```

If it does, I'll close this PR and raise another to revert the asyncio repl change and release this as an App.

ChrisDick:
I've released an App [here](https://github.com/emfcamp/badge-2024-apps-asyncio_repl) to enable the ansyncio repl and #265 has been merged to revert the change that brought it in, so closing this.

PR #262, opened 2026-04-11
added a utility method to guess the address size of eeprom reads/writes

fridgehead:
# Description

Added a function to detect the address length when doing eeprom read and writes, this only triggers for devices with an address of 0x50 (which seems to cover the small 256 byte ones that people use) as I've seen several devices with that address but all using different lengths.

This function has a side effect: it overwrites the first byte of the EEPROM with a '0' value, I've tested this with some of my hexpansions (using variants of the M24C range) and it doesnt seem to affect the badges ability to read the header value.

### DIFF
--- a/modules/system/hexpansion/util.py
+++ b/modules/system/hexpansion/util.py
@@ -5,6 +5,31 @@ from system.hexpansion.header import HexpansionHeader
 
 import typing
 
+def guess_address_length(i2c, addr=0x50):
+    print("Guessing eeprom address length, starting with 2 bytes..")
+    try:
+        i2c.writeto(addr, bytes([0,0]))
+        d = i2c.readfrom(addr, 4)
+        if len(d) == 4:
+            print(".. its a 2 byter!")
+            return 2
+    except:
+        print("..not a 2 byte address, resetting to 1")
+
+    # flush the pending write/read, this overwrites the first byte to 
+    # a 0 which doesnt seem to upset the hexpansion header detection
+    # logic
+    # i've seen this run forever for some reason, so stick some bounds on it
+    ct = 0
+    while ct < 10:
+        try:
+            if i2c.writeto(addr, bytes([0])):
+                break
+            ct += 1
+        except OSError:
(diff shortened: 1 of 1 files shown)


MatthewWilkes:
I don't think it's ideal that this overwrites byte 0, but I can live with it. It would be nice if it could reinstate the 'T', but I think it not making more un-asked for writes is better. We can just live with the magic number changing.

MatthewWilkes:
I don't think I have any of the 256byte eeproms to test, so will leave approving to others

MatthewWilkes:
@fridgehead So, I'm debugging a different eeprom issue right now, and I wonder, do any of these work for your EEPROMs?

1. `HexpansionConfig(3).i2c.readfrom_mem(80, 0, 256)`
2. `HexpansionConfig(3).i2c.readfrom_mem(80, 0, 256, addrsize=8)`
3. `HexpansionConfig(3).i2c.readfrom_mem(80, 0, 256, addrsize=16)`

That's assuming the slot with LEDs 5 and 6, obviously.

fridgehead:
M24C16 -> all 3 functions read correctly
M24C02 -> all 3 functions read correctly (although byte 0 is 0x00 as my changes overwrote it)

Tested this with both this PR and with the last released firmware update, same result

MatthewWilkes:
@fridgehead I am surprised they all work correctly! Would you mind running them each a couple of times in a row? I wouldn't be surprised if some don't reliably read the same offset

MatthewWilkes:
@fridgehead Thanks for this contribution, however I've built the alternate method we tested above into the firmware, so I think it will be working better now. I don't know if guessing the length is still useful (it makes sense it would be), but in the new firmware we're pushing out, it'll be substantially different behaviour. I think we could use this to guess the length in a safer way.

MatthewWilkes:
I'm going to close this for now. @fridgehead if you find a case where the new code path isn't working and we still need it, could you re-open and update? Thanks!

PR #264, opened 2026-04-12
Fix memory leak with images in ctx

### LINKED ISSUE #263: Memory leak in ctx.image()
**Describe the bug**
`ctx.image()` leaks approximately 256KB of RAM per call. Each time a new image is displayed, the previously decoded bitmap is not freed from memory, even after `gc.collect()`. After displaying 6–7 images the device runs out of heap and crashes/resets.

Version: 1.10.0

**To Reproduce**
1. Create a simple carousel app (sample code attached) that displays one of 10 local PNG images (256×256px) on screen.
2. Navigate through images using the left/right buttons — each button press calls ctx.image() with a different path
3. Monitor heap with `gc.mem_free()` before and after each call

Observed heap after each image display:

```
img_0  mem_before=1,883,664  mem_after=1,883,520  (delta=144)
img_1  mem_before=1,621,008  mem_after=1,620,976  (delta=32)
img_2  mem_before=1,359,056  mem_after=1,359,024  (delta=32)
img_3  mem_before=1,096,816  mem_after=1,096,784  (delta=32)
img_4  mem_before=  851,264  mem_after=  851,184  (delta=80)
img_5  mem_before=  589,072  mem_after=  589,040  (delta=32)
→ crash (out of memory)
```

Each image display consumes 262KB (256×256×4 bytes) that is never reclaimed. `gc.collect()` is called explicitly before each load and does not recover it.

**Expected behavior**
Displaying a new image should release the memory used by the previously decoded image. After cycling through multiple images, free heap should remain approximately stable.

**Sample code triggering this issue**
[app.py](https://github.com/user-attachments/files/26646920/app.py)

futureshape:
## Summary

This fixes ctx image texture cache aging on the badge display path.

`display.end_frame()` does not call `ctx_end_frame()`, so ctx's internal texture clock was not advancing between rendered frames. Because texture cache expiry is based on that clock, stale image cache entries could remain alive indefinitely after drawing images with `ctx.image()`.

This means that if an app calls`ctx.image()` multiple times with different images, memory will eventually be exhausted. This fixes #263, verified by building the firmware, flashing the badge and running the sample code from the issue:

```
{'apps.img_test.app.__app_export__': <CarouselApp object at 3c2291c0>}
ImgTest: img_1 size=1247b mem_before=1613488 mem_after=1613456 delta=32
ImgTest: img_2 size=1262b mem_before=1613456 mem_after=1613424 delta=32
ImgTest: img_3 size=1250b mem_before=1613568 mem_after=1613536 delta=32
ImgTest: img_4 size=2945b mem_before=1613680 mem_after=1613600 delta=80
ImgTest: img_5 size=2546b mem_before=1613552 mem_after=1613520 delta=32
ImgTest: img_6 size=1404b mem_before=1613680 mem_after=1613600 delta=80
ImgTest: img_7 size=1408b mem_before=1613472 mem_after=1613440 delta=32
ImgTest: img_8 size=17224b mem_before=1613488 mem_after=1613456 delta=32
ImgTest: img_9 size=16085b mem_before=1613632 mem_after=1613600 delta=32
ImgTest: img_0 size=1261b mem_before=1613568 mem_after=1613536 delta=32
ImgTest: img_1 size=1247b mem_before=1613472 mem_after=1613440 delta=32
ImgTest: img_2 size=1262b mem_before=1613664 mem_after=1613600 delta=64
ImgTest: img_3 size=1250b mem_before=1613472 mem_after=1613440 delta=32
ImgTest: img_4 size=2945b mem_before=1613504 mem_after=1613472 delta=32
```

## Changes

- Advance ctx's texture clock in `tildagon_end_frame()`
- Initialize `reteid` before `ctx_texture_load()` so failed image loads do not read uninitialized stack data (unrelated to the overall issue but probably a good memory safety fix.

## Notes

The decoded STB image buffer is already freed after `ctx_define_texture()`, but ctx keeps its own native texture cache. Advancing the texture clock lets the existing stale-cache logic work as intended.

### DIFF
--- a/components/ctx/ctx.h
+++ b/components/ctx/ctx.h
@@ -55483,7 +55483,7 @@ void ctx_draw_image_clipped (Ctx *ctx, const char *path, float x, float y, float
   else
 #endif
   {
-    char reteid[65];
+    char reteid[65] = "";
     int width, height;
     ctx_texture_load (ctx, path, &width, &height, reteid);
     if (reteid[0])
--- a/drivers/gc9a01/display.c
+++ b/drivers/gc9a01/display.c
@@ -69,10 +69,8 @@ void tildagon_blit_fb (void)
 void tildagon_end_frame(Ctx *ctx)
 {
   ctx_restore (ctx);
+  ctx_end_frame (ctx);
   tildagon_blit_fb ();
-  // display.end_frame() does not call ctx_end_frame(), so advance the
-  // texture clock here to let ctx expire stale image cache entries.
-  ctx_set_textureclock (ctx, ctx_textureclock (ctx) + 1);
   st3m_gfx_fps_update ();
 }
 
--- a/drivers/gc9a01/display.c
+++ b/drivers/gc9a01/display.c
@@ -70,6 +70,9 @@ void tildagon_end_frame(Ctx *ctx)
 {
   ctx_restore (ctx);
   tildagon_blit_fb ();
+  // display.end_frame() does not call ctx_end_frame(), so advance the
+  // texture clock here to let ctx expire stale image cache entries.
+  ctx_set_textureclock (ctx, ctx_textureclock (ctx) + 1);
   st3m_gfx_fps_update ();
 }
 
--- a/drivers/gc9a01/display.c
+++ b/drivers/gc9a01/display.c
@@ -69,8 +69,11 @@ void tildagon_blit_fb (void)
 void tildagon_end_frame(Ctx *ctx)
 {
   ctx_restore (ctx);
-  ctx_end_frame (ctx);
   tildagon_blit_fb ();
+  // display.end_frame() cannot call ctx_end_frame() directly here: that resets
+  // rasterizer state, including the framebuffer clip bounds, which leaves
+  // subsequent frames blank. Advance only the texture eviction clock.
+  ctx_set_textureclock (ctx, ctx_textureclock (ctx) + 1);
   st3m_gfx_fps_update ();
 }
 


MatthewWilkes:
Does it make sense for us to add a call to `ctx_end_frame` instead of manually manipulating the texture clock?

futureshape:
Possibly - but I'm not super familiar with ctx and I don't know if the `ctx_end_frame` omission was deliberate (because it has some side effects?) or accidental. 

I'll mark it as draft for now and test with `ctx_end_frame`

MatthewWilkes:
It was probably accidental. I suspect there will be side effects to calling end_frame, but if it also fixes the problem, then I'd be inclined to go with it. Those side effects might be other latent bug fixes 🙃

futureshape:
@MatthewWilkes FYI, I tried both:

```
void tildagon_end_frame(Ctx *ctx)
{
  ctx_restore (ctx);
  ctx_end_frame (ctx);
  tildagon_blit_fb ();
...
```

and a different order, just in case

```
void tildagon_end_frame(Ctx *ctx)
{
  ctx_restore (ctx);
  tildagon_blit_fb ();
  ctx_end_frame (ctx);
...
```

but both have the same result of breaking the display 😭 

1. The badge boots OK
2. The initial menu is shown momentarily and then disappears
3. Connecting via mpremote, I can see that the issue is limited to the display, the badge has not crashed, if I press up/down and then select I can see in the output that different apps are loading, and back returns to the launcher app

I will do a bit more investigation on what else `ctx_end_frame` does to see if I can understand what's causing this, but we may need to revert to the more hacky fix.

MatthewWilkes:
Let's just do the hacky fix. If you want to bring this PR back I will test and merge. Thanks!

futureshape:
Done, I've taken it out of draft. I've also tested with a couple of apps that do continuous drawing (Bubble and Starfield) and it seems ok.

MatthewWilkes (approved):
Thanks for this!

PR #265, opened 2026-04-12
revert asyncio REPL to move it to an app

ChrisDick:
# Description

asyncio REPL breaks Thonny and other desktop apps. I've written an app to enable the asyncio REPL instead of adding another setting.


MatthewWilkes (approved):
Looks good

PR #266, opened 2026-04-12
Improve error handling for Hexpansion insertion

MatthewWilkes:
# Description

This improves the Hexpansion insertion experience by catching OSErrors during Hexpansion enumeration, and cancelling the insertion with an error instead. This won't trigger for electronic hexpansions without an EEPROM, only for ones that have a detected EEPROM where the header can't be read.

### DIFF
--- a/modules/system/hexpansion/app.py
+++ b/modules/system/hexpansion/app.py
@@ -234,9 +234,22 @@ class HexpansionManagerApp(app.App):
             return
 
         # Do we have a header?
-        header = read_hexpansion_header(i2c, addr, addr_len=addr_len)
-        if header is None:
+        try:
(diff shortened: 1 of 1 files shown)


Jonty:
Is this related? https://github.com/emfcamp/badge-2024-software/pull/190

MatthewWilkes:
Not really, they're in a similar code path, but quite different. That one allows failed reads to be ignored, which can result in partial reads of the filesystem but prevents crashes. This prevents errors on hexpansion insertion if there's a i2c bus conflict or an unprovisioned EEPROM, and shows a user-facing warning. Previously, this would either be invisible or would crash the badge. Failed reads/writes during interactions still cause crashes, this only protects against successful reads of a bad hexpansion.

ChrisDick (approved):

PR #267, opened 2026-04-12
Use i2c readfrom_mem family of APIs for reading from EEPROMs

MatthewWilkes:
Currently, we write addresses then read using the general read/write methods on i2c. This has worked for lots of chips, but I have a Zetta EEPROM which is failing with this. I suspect a timing issue. I've switched the API into the mem variants, and the EEPROMs I have are working correctly, including the new one.

### DIFF
--- a/modules/system/hexpansion/header.py
+++ b/modules/system/hexpansion/header.py
@@ -128,7 +128,7 @@ def write_header(port, header, addr=0x50, addr_len=2, page_size=32):
 def read_header(port, addr=0x50, addr_len=2):
     i2c = I2C(port)
 
-    header_bytes = i2c.readfrom_mem(addr, 0, 32, addr_bytes=addr_len * 8)
+    header_bytes = i2c.readfrom_mem(addr, 0, 32, addrsize=addr_len * 8)
     header = HexpansionHeader.from_bytes(header_bytes)
 
     return header
--- a/modules/system/hexpansion/header.py
+++ b/modules/system/hexpansion/header.py
@@ -128,11 +128,7 @@ def write_header(port, header, addr=0x50, addr_len=2, page_size=32):
 def read_header(port, addr=0x50, addr_len=2):
     i2c = I2C(port)
 
-    # Set internal address to 0x00
-    addr_bytes = [0] * addr_len
-    i2c.writeto(addr, bytes(addr_bytes))
-
-    header_bytes = i2c.readfrom(addr, 32)
+    header_bytes = i2c.readfrom_mem(addr, 0, 32, addr_bytes=addr_len * 8)
     header = HexpansionHeader.from_bytes(header_bytes)
 
     return header
--- a/modules/write_header.py
+++ b/modules/write_header.py
@@ -23,10 +23,7 @@ def write_header(port, addr=0x50):
 def read_header(port, addr=0x50):
     i2c = I2C(port)
 
-    # Set internal address to 0x00
-    i2c.writeto(addr, bytes([0, 0]))
-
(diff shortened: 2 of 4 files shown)


modules/system/hexpansion/header.py
```diff
@@ -128,11 +128,7 @@ def write_header(port, header, addr=0x50, addr_len=2, page_size=32):
 def read_header(port, addr=0x50, addr_len=2):
     i2c = I2C(port)
 
-    # Set internal address to 0x00
-    addr_bytes = [0] * addr_len
-    i2c.writeto(addr, bytes(addr_bytes))
-
-    header_bytes = i2c.readfrom(addr, 32)
+    header_bytes = i2c.readfrom_mem(addr, 0, 32, addr_bytes=addr_len * 8)
```
Robotmad: I think addr_bytes should be addrsize? default for which is 8 so this will be working with small EEPROMs.

modules/write_header.py
```diff
@@ -23,10 +23,7 @@ def write_header(port, addr=0x50):
 def read_header(port, addr=0x50):
     i2c = I2C(port)
 
-    # Set internal address to 0x00
-    i2c.writeto(addr, bytes([0, 0]))
-
-    header_bytes = i2c.readfrom(addr, 32)
+    header_bytes = i2c.readfrom_mem(addr, 0, 32)
```
Robotmad: Believe it needs addrsize setting here in order to work with all EEPROMs, hence can only be called once you have worked that out.

fridgehead: I tested these methods with a 256 byte eeprom (that uses 8 bit addressing) and it didnt seem to matter what the `addr_bytes` value was set to, in all cases it read the eeprom metadata fine.

I've just flashed the `wilkes/eeprom-addressing-memfunction` branch and its detecting my 8byte address eeproms:

`
Hexpansion inserted in port: 2
Found hexpansion header:
HexpansionHeader[
    manifest version: 2024,
    fs offset: 32,
    eeprom page size: 16,
    eeprom total size: 2048,
    vendor id: 0xBEAF,
    product id: 0xDEED,
    unique id: 0,
    friendly name: DECTalk,
]
Warning: possible unsupported chip. Size: 2048
1 chips detected. Total EEPROM size 2048bytes.
Page size: 16
eeprom block count: 32
partition block count: 31
partition block size: 64
Attempting to mount i2c eeprom from hexpansion port 2
Failed to mount: [Errno 5] EIO
`

Robotmad: Umm?  If I call readfrom_mem() with "addr_bytes=" I get an exception "extra keyword arguments given" which I think is because, as per the edit in util.py the parameter is actually "addrsize".    Need to trace through which version of read_header() is actually being used where, as it exists in modules/write_header.py, modules/system/hexpansion/header.py and as read_hexpansion_header() in modules/system/hexpansion/util.py.  I think it is actually this last one (correctly edited in this commit) which is called from handle_hexpansion_insertion() which generating your log message.  So I think that proves it will work, but needs to use addrsize and not addr_bytes.

modules/system/hexpansion/header.py
MatthewWilkes: Yes, you're right. I'll fix that.

modules/write_header.py
MatthewWilkes: @Robotmad You're right, it's just the default of 16 seems to work pretty reliably with all eeproms currently in use. I'll fix those.

MatthewWilkes:
@ChrisDick Addressed those bugs

ChrisDick (approved):

PR #271, opened 2026-04-16
Remove dup unused file

garthy:
# Description

Removes a duplicate file that is not use here.

The actual used  file is `components/flow3r_bsp/flow3r_bsp_display.c`

This confused me when trying to understand the code.


ChrisDick (commented):
the only difference between the two files is the spi host, so this one can go

ChrisDick (approved):

PR #273, opened 2026-04-17
Remove unused gc9a01py.py file (Saves 8976 Bytes)

garthy:
# Description

Looks like this python interface to the Display is never really used. Removing saves 8976 Bytes

Also reduces confusion when trying to understand the code. 

I was starting to look at replacing this with the `C` version of this code when I realised this wasn't used.

## Before
```
bootloader  @0x000000    19440  (   13328 remaining)
partitions  @0x008000     3072  (    1024 remaining)
application @0x010000  2262704  (  293200 remaining)
total                  2328240
```
## After
```
bootloader  @0x000000    19440  (   13328 remaining)
partitions  @0x008000     3072  (    1024 remaining)
application @0x010000  2253728  (  302176 remaining)
total                  2319264
```


ChrisDick:
unfortunately tildagonos.displayinit is used by [NottingHackApp ](https://github.com/aaronjackson/tildagon-nottinghack-logo) and [Bitmap-Demo-main](https://github.com/ntflix/Tildagon-Bitmap-Demo), we would have to do something to keep them working. would be good to get those bytes back.

AaronJackson:
I don't mind if you break it :sunglasses:

Jonty:
We can just make it a noop I think, but we can also just PR updates to those apps :D

ChrisDick:
they'd need to be changed from bitmaps to jpg or svg for ctx

MatthewWilkes (approved):
Given we're gearing up for a v2, I think we should just merge this. It will break those apps, but this is the time to make breaking changes, and it sounds like its not super controversial.

Issue #269, opened 2026-04-13
Simulator crashes when drawing images since #240


mbooth101:
**Describe the bug**

The simulator crashes when I try to use `ctx.image()` in my app, I get the following:

```
Traceback (most recent call last):                                                                                                                                                                                    
  File "/home/mbooth/emfcamp/badge-2024-software/modules/system/scheduler/__init__.py", line 221, in _render_task                                                                                                     
    app.draw(ctx)                                                                                                                                                                                                     
  File "/home/mbooth/emfcamp/badge-2024-software/sim/apps/mbooth101_emf_settlers/app.py", line 20, in draw                                                                                                            
    ctx.image(TITLE_IMG, -120, -45, 240, 90)                                                                                                                                                                          
  File "/home/mbooth/emfcamp/badge-2024-software/sim/fakes/ctx.py", line 409, in image                                                                                                                                
    _img_cache[path] = _wasm.stbi_load_from_memory(buf)                                                                                                                                                               
  File "/home/mbooth/emfcamp/badge-2024-software/sim/fakes/ctx.py", line 166, in stbi_load_from_memory                                                                                                                
    b = mem[res:]                                                                                                                                                                                                     
ValueError: slice stop is required
```

Version: 4fce01000f995ee3d62b3646374e7c09d7476bfd

**To Reproduce**

Write a trivial app to render an image, e.g:

```
import app
import os
                                                                                                                                                                                                                      
TITLE_IMG = "/apps/sim_crash/title.png"
if os.getcwd() != "/": 
    TITLE_IMG = os.getcwd() + TITLE_IMG

class SimCrash(app.App):

    def __init__(self):
        pass

    def update(self, delta):
        pass

    def draw(self, ctx):
        ctx.rgb(0,0,0).rectangle(-120, -120, 240, 240).fill()
        ctx.image(TITLE_IMG, -120, -45, 240, 90)                                                                                                                                                                      
                                                                                                                                                                                                                      
__app_export__ = SimCrash
```

Where title.png is some small image: https://github.com/mbooth101/emf-settlers/blob/main/title.png

The code in the stacktrace was added in PR #240

mbooth101:
I submitted a PR to fix this in #275

mbooth101:
Thanks for merging my PR

PR #275, opened 2026-04-24
Fix simulator crash when drawing images (#269)

mbooth101:
A Tildagon app rendering an image may encounter "ValueError: slice stop is required"

The wasm runtime doesn't permit unbounded slicing of its memory buffers, so just calculate the size we expect the buffer to be and use that as the stop index


Funcan (approved):

Funcan (approved):
Looks entirely valid to me, and only affects the simulator

PR #280, opened 2026-05-02
Add Hexpansion firmware manager app

MatthewWilkes:
This is a draft for a new built-in app that allows updating hexpansions with an EEPROM. It also allows for factory resetting, and bulk provisioning for hexpansion creators.

This will need the repo repointing to EMF, and I want to make it configurable where it looks, and have bulk provisioning be hidden behind a developer mode toggle

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,6 +28,9 @@ jobs:
       - name: Build firmware
         run: |
           docker run -q --rm --env "TARGET=esp32s3" -v "$(pwd)"/:/firmware ghcr.io/emfcamp/esp_idf:v5.5.1
+      - name: Build mpy-cross image
+        run: |
+          docker build -f Dockerfile.mpcross --build-arg IDF_VERSION=v5.5.1 -t ghcr.io/emfcamp/mpy-cross:v5.5.1 .
       - name: Archive firmware
         uses: actions/upload-artifact@v4
         with:
--- a/Dockerfile.mpcross
+++ b/Dockerfile.mpcross
@@ -1,0 +1,6 @@
+ARG IDF_VERSION=v5.2.1
+FROM ubuntu:25.10
+#FROM ghcr.io/emfcamp/esp_idf:$IDF_VERSION
+
+COPY micropython/mpy-cross /mpy-cross
+ENTRYPOINT ["/mpy-cross/build/mpy-cross"]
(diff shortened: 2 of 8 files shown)


thinkl33t:
Seems to work nicely.

- Might be worth chucking the short git hash of the latest release into the eeprom fs if there's room, so we can easily check if the eeprom needs updating and show the hash of the currently deployed version
- There isn't much feedback that the updater is actually doing anything in between pressing 'update' and the 'updated' popup - maybe a new screen with an arrow that does a 'downloading / flashing / done' status
- Can we unmount / remount the hexpansion when an update is performed so the new code is loaded?

thinkl33t:
How does bulk provisioning mode work?  Is it storing the max number flashed for an vid/pid pair on the badge or does it go from the number of the currently plugged in hexpansion?

thinkl33t:
I'm unclear what factory reset *does* - does it wipe the eeprom and install the latest firmware?

MatthewWilkes:
1.  Might be worth chucking the short git hash of the latest release into the eeprom fs if there's room, so we can easily check if the eeprom needs updating and show the hash of the currently deployed version **I was wondering about this, it would need another HTTP round-trip to get metadata when updating, but it is very doable. The thing I've been unsure about for this is that it'll use up some extra bytes on the EEPROM, and it's a big struggle to fit things in sometimes. That's why I went with just having people try and update whenever**
2. There isn't much feedback that the updater is actually doing anything in between pressing 'update' and the 'updated' popup - maybe a new screen with an arrow that does a 'downloading / flashing / done' status **Good idea**
3. Can we unmount / remount the hexpansion when an update is performed so the new code is loaded? **Hmm, good idea.**
4. How does bulk provisioning mode work? Is it storing the max number flashed for an vid/pid pair on the badge or does it go from the number of the currently plugged in hexpansion? **So, it's janky. It takes the one of the one that's plugged in, then increments it. So, if you're doing it in multiple batches, it assumes you start with the latest one plugged in. This should get changed, I agree**

5. I'm unclear what factory reset does - does it wipe the eeprom and install the latest firmware? **Yup**

@thinkl33t Is it good enough to go into the alpha, do you think, or do these need to get fixed first?

sodoku:
I tested this with a hexpansion that is registered, but does not have an app.py in the hexpansion-firmwares repo.

- The hexpansion only shows up in the app, when I insert the hexpansion before running the app. If I run the app with an inserted hexpansion then remove it and open the app again, it still lists the hexpansion. Is the app correctly handling insertion and removal of hexpansion?

- I can run "update" and nothing happens, "bulk provision" and "factory reset" both crash the app. I assume because it tries to download "https://github.com/emfcamp/hexpansion-firmwares/releases/download/latest//firmware_0xBAD3_0x4EEB.tar.gz" which does not exist.

- The app allows me to configure the 2024 front board. Should I be able to do this? If yes, the firmware should probably exist to not crash the app (see above)

MatthewWilkes:
@thinkl33t @sodoku Addressed all of those points except the git hash

MatthewWilkes:
@thinkl33t Given we are in JFDI, can we merge this and continue improving the bulk updates async?

thinkl33t:
Aye let's go for it

PR #283, opened 2026-05-04
Upgrade to Micropython 1.28

MatthewWilkes:
# Description

Building on top of @Molive-0's micropython 1.25 PR (#200), this further upgrades to Micropython 1.28 on ESP-IDF 5.5.1. 

## Changes

* Change WPA Enterprise to a new API version
* Intern strings in MicroPython exceptions raised in C

## Todo
* The USB support has been updated in this version of Micropython, I've not dug into this yet (although I think @Jonty may have had a play?)

## Current status
* Builds in Docker, and flashes to the badge. Smoke test passes.


Jonty:
I dug into a bunch of USB stuff before realising that the newer versions had totally rewritten it, so binned the lot and decided to do the upgrade instead, which I never finished!

MatthewWilkes:
FYI, the build check here cannot pass until the docker image is built, but that only happens on merge.

thinkl33t:
I can build and push something into the registry so we can get past the initial hurdle if we want

MatthewWilkes:
@thinkl33t Ooh, yes please. It's just the standard build, but with the new 5.5.1 version in the arg

thinkl33t:
Done

thinkl33t (approved):
Builds, that means it works 😬

Issue #284, opened 2026-05-05
MicroPython soft reset breaks front board buttons


b0xcat:
**Describe the bug**
The 2024 frontboard ePin button interrupt handlers stop firing after a MicroPython soft reset (e.g. when using mpremote mount). Buttons work fine on a normal boot.

`machine_pins_deinit()` is called during every soft reset, which removes all ISR handlers, including the GPIO 10 one for the AW9523B GPIO expanders (and USB stuff).

`tildagon_power_fast_task` registers the GPIO_NUM_10 ISR handler only once during task creation, i.e. only on hard boot. After soft reset removes the handler, it is never re-registered. The power task doesn't get INTERRUPT_EVENTs anymore, and the buttons stop working. This might also have implications for the USB system, but I'm not really familiar with how that works.

Version: [current main, e8190290f1bae7bcfedad0ccd55611ca7b87b95f at the time of writing]

**To Reproduce**
1. Boot the badge normally (buttons work)
2. Connect via mpremote mount .
3. Run main.py from the mounted filesystem
4. Buttons no longer respond

**Expected behavior**
Buttons keep working, even during repl use

ChrisDick:
fixed in #285

PR #285, opened 2026-05-05
Re-register GPIO 10 isr handler after soft reset

b0xcat:
# Description

This PR introduces a way to automatically re-register the isr handler for GPIO 10 in `tildagon_power_fast_task`, which gets de-registered on soft reset, which breaks the dev workflow of doing `mpremote mount... etc` (see #284)

It introduces a timeout for the `xQueueReceive` in `tildagon_power_fast_task`, and if it times out and GPIO 10 got stuck indicating an unhandled interrupt, it re-registers the handler. This feels kinda hacky and maybe fragile.

If anyone knows a better way of fixing this, I'm 100% for it. Just let me know!

### DIFF
--- a/drivers/tildagon_power/tildagon_power.c
+++ b/drivers/tildagon_power/tildagon_power.c
@@ -141,7 +141,7 @@ void tildagon_power_fast_task(void *param __attribute__((__unused__)))
         while ( 1 )
         {
             event_t event = NO_EVENT;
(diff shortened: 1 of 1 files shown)


ChrisDick (approved):
This will help it if the isr gets stuck for any reason. Shame there's no to check the isr is already registered, but no harm in re-registering it if it's not the repl that caused the hang.

PR #268, opened 2026-04-13
Force NiMBLE to use the external spi ram

Jonty:
Hopefully this will resolve the memory issues when enabling bluetooth, however I do not have a tildagon to hand to test it yet, and I am writing this on my phone from a train. Given the default mode should take spiram into account this may do nothing at all, but I have found reports elsewhere that it worked for them.

Testing this should be as simple as `import bluetooth; ble = bluetooth.BLE(); ble.active(True)` - if the badge doesn't crash it worked.

Ref https://github.com/emfcamp/badge-2024-software/issues/13

Issue #13, opened 2024-05-02
Enable Bluetooth support


robmckinnon:
Can we please consider enabling Bluetooth support in the MicroPython firmware for the badge by default?

What are the arguments for or against this? 

Is it a power consumption issue, is there a way around that?

I'd like to write an app for the badge to use [BLE MIDI](https://midi.org/category/midi-connections/bluetooth-low-energy-ble-wireless-midi) to communicate wirelessly with BLE MIDI enabled synth and MIDI controller hardware. I imagine this would be easier to develop and share with others if Bluetooth was enabled by default.

In the [mpconfigboard.h](https://github.com/search?q=repo%3Aemfcamp%2Fbadge-2024-software%20mpconfigboard.h&type=code) file adding `#define MICROPY_PY_BLUETOOTH                (1)` enables Bluetooth support in the MicroPython firmware.

MatthewWilkes:
@robmckinnon BLE support is already enabled, but the full Bluetooth stack isn't. Is that sufficient for you?

At https://github.com/emfcamp/badge-2024-software/blob/main/tildagon/mpconfigboard.cmake#L39 we include the following config file: https://github.com/micropython/micropython/blob/cfd5a8ea3a80930a79dbae6a6aff54587393d0b8/ports/esp32/boards/sdkconfig.ble

I can confirm that `bluetooth.BLE` is available on our test builds, and we've been able to view advertisements.  Does that cover your needs?

fridgehead:
yeah it works for us too, using aioble to talk to a strip of LED's in our dome:
```

async def connectDevice(self):
        device = aioble.Device(aioble.ADDR_PUBLIC, "23:00:00:00:00:00")
        conn = await device.connect()
        try:
            connection = await device.connect(timeout_ms=2000)
        except asyncio.TimeoutError:
            print('Timeout')
        print("connected!")
        temp_service = await connection.service(bluetooth.UUID(0xafd0))
        temp_char = await temp_service.characteristic(bluetooth.UUID(0xafd1))
        print("offen")
        offbytes = b"\x5B\x0f\x00\xB5"
        await temp_char.write(offbytes, timeout_ms=1000)

```

augeas:
BLE still seems somewhat dead:

This keels over with `OSError: [Errno 5] Input/output error` in the REPL...

```python

import aioble
import app``
import asyncio

class TildaMesh(app.App):
    def __init__(self):
        super().__init__()
        self.devices = []

[...]

PR #287, opened 2026-05-05
Move CTX framebuffer and nimble to PSRAM

MatthewWilkes:
# Description

Refs #268 and #13 - moves the ctx framebuffer and nimble memory to PSRAM, which re-enables BLE. Note, ea515cbca057f666dbc70a2e1939cc1b6aeb7390 talks about moving out of PSRAM due to cache line conflicts, but it seems to work without this fix, so my current guess is that this was part of the CTX image loading debugging, and was a red herring


thinkl33t (approved):
LGBTM

PR #290, opened 2026-05-06
Sim: allow faked sys.print_exception to take multiple arguments

### LINKED ISSUE #289: Sim hangs with no error when exception happens in update()
**Describe the bug**

I noticed that I'd sometimes get hangs in the simulator, it would freeze with no error. But if I ran the same code on the device I'd get a useful stack trace. This limits the ability to debug on the sim.

This happens if there is an exception raised in the `update()` method of an app. Exceptions in `__init__()` or `draw()` do get reported properly in the sim.

Version: Current main at v2.0.0-alpha.2 (but also with a checkout from a few days ago)

**To Reproduce**

Modify the example app (`sim/apps/example/app.py`)
```python
import asyncio
import app


class ExampleApp(app.App):
    def __init__(self):
        print("__init__")

    def update(self, delta):
            self.minimise()
        print("update")
        raise RuntimeError

    def draw(self, ctx):
        ctx.save()
        ctx.rgb(0.2,0,0).rectangle(-120,-120,240,240).fill()
        ctx.rgb(1,0,0).move_to(-80,0).text("Hello world")
        ctx.restore()
        print("draw")
```
Launch with `run.py example.ExampleApp`

**Expected behavior**
A back trace from `update()` showing where the raise happens.

On the device I get
```
  File "system/scheduler/__init__.py", line 181, in app_wrapper
  File "app.py", line 21, in run
  File "apps/example/app.py", line 15, in update
RuntimeError: 
```

**Actual behaviour**
I see
```
__init__
Registered async event handler for ShowNotificationEvent: NotificationService - _handle_incoming_notification
update
draw
```
so it looks like the crash in `update()` does not prevent `draw()` running. But it fails to run update again.

**Additional context**

samtygier:
The sim creates a fake `sys.print_exception` to match micropython. Its actual arguments are ignored, it will always print the current trace. In some places it is called with 2 arguments, so allow it to handle an arbitary number.

This allows stack traces to be printed correctly in some catch blocks when run in the sim.

Closes #289

### DIFF
--- a/sim/run.py
+++ b/sim/run.py
@@ -143,7 +143,7 @@ def mkstat(orig_stat):
 os.stat = mkstat(os.stat)
 
 
-sys.print_exception = lambda x: print(traceback.format_exc())
+sys.print_exception = lambda *_: print(traceback.format_exc())
 
 
 def replace_launcher(module_name: str, class_name: str):


samtygier:
Alternative solutions
*  Handle the `file` argument and pass that to `print`. So that you can choose to print the back trace to `stderr`
* Remove the file arguments from `sys.print_exception` . I am pretty sure `mpremote` merges all output to stdout.

ChrisDick (approved):

PR #291, opened 2026-05-09
Derive app_module_name from metadata

pikesley:
# Derive app_module_name from metadata

```
```

As far as I can tell this is exacly what we want, but this may be very naive on my part (attempting to work out what's happening where this code is digging through the tarball to derive the path just left me confused).

## Note

~As far as I can tell, the only difference between the tarballs from the different forges is that a Github tar unpacks to a directory with the release name appended:~

No, it's actually [way stupider](https://github.com/emfcamp/badge-2024-software/pull/293)

```
tildagon-acapp-0.0.3
```

whereas a Codeberg one does not:

```
tildagon-lemmings
```

So I cannot work out *why* it's generating something so broken

### DIFF
--- a/modules/firmware_apps/app_store.py
+++ b/modules/firmware_apps/app_store.py
@@ -611,14 +611,6 @@ def install_app(app):
         raise e
 
 
-# I don't believe this is ever called
-# def validate_app_files(tar):
-#     prefix = find_app_root_dir(tar)
-#     app_py_path = find_app_py_file(prefix, tar)
-#     print(f"Found app.py at: {app_py_path}")
-#     return prefix
-
-
 def find_app_root_dir(tar):
     print("Finding root dir...")
     root_dir = None
--- a/modules/firmware_apps/app_store.py
+++ b/modules/firmware_apps/app_store.py
@@ -558,7 +558,14 @@ def install_app(app):
         except OSError:
             pass
 
-        app_module_name = "_".join(prefix.split("-")[0:-1])
+        # I think this is all we need
+        # So like
+        # https://whatever.forge/dave-user/tildagon-some-app
+        # resolves to
+        # dave_user_tildagon_some_app
+        app_module_name = "_".join([app["id"]["owner"], app["id"]["title"]]).replace(
+            "-", "_"
+        )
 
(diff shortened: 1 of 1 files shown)


b0xcat:
Best I can tell the current app store code attempts to strip off the version number with 
```python
app_module_name = "_".join(prefix.split("-")[0:-1])
```
so if that's not there, as it seems to for codeberg apps, it just strips off the last bit after -

For example, `tildagon-particle-man` incorrectly gets installed to `tildagon-particle`

Grabbing this info from the app store index json instead seems reasonable to me, although I'm not sure about the details of how that's generated and if it results in the same path for at least all github apps.

LGTM, but maybe double check with @hughrawlinson to be sure.

b0xcat:
Actually sorry I guess we don't currently prepend the owner to the path, which might break apps that import their own files as `apps.appname.something`

pikesley:
> Best I can tell the current app store code attempts to strip off the version number with
> 
> ```python
> app_module_name = "_".join(prefix.split("-")[0:-1])
> ```
> 
> so if that's not there, as it seems to for codeberg apps, it just strips off the last bit after -
> 
> For example, `tildagon-particle-man` incorrectly gets installed to `tildagon-particle`
> 
> Grabbing this info from the app store index json instead seems reasonable to me, although I'm not sure about the details of how that's generated and if it results in the same path for at least all github apps.
> 
> LGTM, but maybe double check with @hughrawlinson to be sure.

Oh wait this is a handy clue 👍

pikesley:
If I bypass the attempt to strip the version number for a Codeberg app, then it gets _slightly_ better:

```
Found app prefix: tildagon-barnsley-fern
```

pikesley:
I have an [alternative approach](https://github.com/emfcamp/badge-2024-software/pull/293)

MatthewWilkes (approved):
Agree on getting this working. I like this approach over the other. It doesn't encode behaviour of third parties, it just sets it. If we weren't in JFDI time I would suggest deleting code rather than commenting out, but this is fine.

You may also want to PR it into rc-1.x, depending on if we want to release it sooner than 2.x

pikesley:
SUperseded by #301

PR #293, opened 2026-05-10
Construct the `app_module_name` based on the source service

pikesley:
# Derive app_module_name based on source service

A possibly safer (but maybe less elegant) approach than #291 

The problem is all in the tarballs. When we unpack a Github download we get something like:

> Note: what we get from downloading via the API is _not_ the same thing we get from the "Download releases" webpage. Thanks, Github

`tildagon-my-app`

No owner-name, and no trailing hash. This PR adds a little helper method that prepends the ownername if we're coming from Codeberg, and only strips the trailing hash if we're coming from Github

If we like this approach, I can clean it up a little and maybe write some tests

### DIFF
--- a/modules/firmware_apps/app_store.py
+++ b/modules/firmware_apps/app_store.py
@@ -558,7 +558,7 @@ def install_app(app):
         except OSError:
             pass
 
-        app_module_name = "_".join(prefix.split("-")[0:-1])
+        app_module_name = derive_app_module_name(prefix, app)
 
         t = TarFile(fileobj=tar_bytesio)
         for i in t:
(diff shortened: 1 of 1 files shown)


modules/firmware_apps/app_store.py
```diff
@@ -604,11 +604,33 @@ def install_app(app):
...
+def derive_app_module_name(prefix, metadata):
+    # with a Github tarball (the default):
+    # the owner name is baked into the root directory
+    owner = ""
+
+    # and a hash is attached to the end, which we want to strip
+    final_index = -1
+
+    # with a Codeberg tarball:
+    if metadata["id"]["service"] == "codeberg":
```
hughrawlinson, line 616: I don't think we should be doing different things conditionally on the code forge if we can get away with it - I prefer your original solution I think, creating the metadata from the app store data.

pikesley: Agreed, I was just concerned that there were more subtle things going on here that I was not aware of

pikesley:
Superseded by #301

PR #294, opened 2026-05-10
Utilities for composing NeoPixels

MatthewWilkes:
This adds some classes that allow complex composition of NeoPixels, including:

1. MergedNeoPixel - allows multiple LEDs to act as a single item in the chain
2. ComposedNeoPixel - allows multiple chains to be controlled as if they were one, including with overlaps
3. CorrectedNeoPixel - allows for values assigned to items in chains to be modified in-place during set

### DIFF
--- a/modules/system/patterndisplay/app.py
+++ b/modules/system/patterndisplay/app.py
@@ -4,6 +4,7 @@ import settings
 import asyncio
 import os
 import sys
+import neopixel
 from system.patterndisplay.events import (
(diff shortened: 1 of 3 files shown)


ChrisDick (approved):
does this change mean that we need to re run firstTime.sh?

MatthewWilkes:
@ChrisDick It will, yes. Without that, the patterns would be broken.

PR #297, opened 2026-05-10
Added lieff/minimp3 library to decode mp3 audio

andreacampanella:
# Description

Adds minimp3, an header only mp3 decoder (https://github.com/lieff/minimp3)

This is been tested with  https://github.com/andreacampanella/BadgeRadio

### DIFF
--- a/drivers/micropython.cmake
+++ b/drivers/micropython.cmake
@@ -29,3 +29,4 @@ include(${CMAKE_CURRENT_LIST_DIR}/tildagon/micropython.cmake)
 
 # Add burnt-in HMAC
 include(${CMAKE_CURRENT_LIST_DIR}/tildagon_hmac/micropython.cmake)
+include(${CMAKE_CURRENT_LIST_DIR}/mp3/micropython.cmake)
--- a/drivers/mp3/micropython.cmake
+++ b/drivers/mp3/micropython.cmake
@@ -1,0 +1,21 @@
+add_library(usermod_mp3 INTERFACE)
+
+target_sources(usermod_mp3 INTERFACE
+    ${CMAKE_CURRENT_LIST_DIR}/mp3module.c
+)
+
+target_include_directories(usermod_mp3 INTERFACE
+    ${CMAKE_CURRENT_LIST_DIR}
+)
+
+target_compile_options(usermod_mp3 INTERFACE
+    -Wno-unused-function
+    -Wno-unused-variable
+    -Wno-unused-but-set-variable
+    -Wno-shift-negative-value
+    -Wno-implicit-fallthrough
+    -Wno-double-promotion
+    -Wno-float-conversion
+)
+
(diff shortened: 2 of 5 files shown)


MatthewWilkes:
@andreacampanella This looks like a substantial dependency, how much does it eat of the firmware size budget? Have you considered the tradeoffs of compiling it to an mpy file and bundling with the app?

andreacampanella:
I spent the last 2 weeks trying to integrate the code in a module but I hit what looks like a bug in the xtensa SDK that I cannot figure out, the code takes 37k in ROM and 7k in ram, I would love if someone with more experience then me in micropython can figure out why the module doesn't work, but without JTAG access to the badge I had a very hard time trying.

kliment:
The JTAG pins are brought out to one of the hexpansion headers but if you want to try debugging it's probably easiest using an espressif S3 devboard

MatthewWilkes:
@andreacampanella do you have any info about the problems you were seeing? It sounds like a fun project, and of course nobody wants projects to fail at the last hurdle, but given the badge doesn't have sound hardware, I think understanding why this would need to be special cased and if/how it could migrate to a native module in future (and what bug blocks that) would be helpful. Otherwise, we would be somewhat at risk of the firmware bloating.

andreacampanella:
I wish I had more information, but basically, what happens is that I can build the module, and load it, but when i try to init it, the badge just restart, no output,nothing , I thought it was a problem with my c code but an example loaded well, and the same module loads in a container with micropython running on x86, I seem to be limited to one MP_DEFINE_CONST_FUN_OBJ_0 , if there is a known working implementation of how to load a c library on the badge codebase that would be very helpful but I didn't spot any. 

About the sound hardware, in the next days I was going to design and release an hexpansion to build one, it would have been cheap, as the DAC is off the shelf and wires directly to the port, no other components needed (a video can be found here: https://bsky.app/profile/emuboy.bsky.social/post/3mlim63ajxs24 ) 

I totally understand if this is not the way to go, I was just hoping I didn't had to go full dive on it...

andreacampanella:
Following kliment''s advice (thanks!) on using an dev board I did managed to get the minimp3 code working isnide the app, but I'm now facing a very weird problem, my app seems to stress both wifi and cpu a LOT, that's without the probably horrible EMFs generated by the I2S interface, anyway, a couple of time, after some minutes of running, the badge crashed and I couldn't start the stream again until I pulled the battery, even after multiple soft reboot.

is there a place were badge developers hangs out? because I have questions :D

kliment:
I suspect you're running low on RAM and it works fine on the devkit that isn't running the whole display framework which uses a lot of RAM, but fails on the badge when it can no longer allocate enough memory to do its thing. There is a badge channel on IRC `#emfcamp-badge` on liberachat and Matrix `#badge:emfcamp.org` (they are bridged together) if you have further questions.

PR #300, opened 2026-05-13
Fix previously installed apps not launching when using mpremote mount

b0xcat:
# Description

This PR fixes a small issue launching apps that were previously already installed when using `mpremote mount`.

`mpremote mount` under the hood does `os.chdir('/remote')` which causes apps previously installed to `/apps` to fail when we try to import them. This provides a fallback that temporarily switches back to the old cwd, imports the app, and restores the /remote cwd after.

### DIFF
--- a/modules/system/launcher/app.py
+++ b/modules/system/launcher/app.py
@@ -157,7 +157,9 @@ class Launcher(App):
             except ImportError as e:
                 cwd = os.getcwd()
                 if "no module named 'apps'" in str(e) and cwd == "/remote":
-                    print("Failed to launch app from mpremote mount, launching from internal storage instead")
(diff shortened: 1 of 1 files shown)


MatthewWilkes:
@b0xcat Won't lots of things be broken if `/remote` is in place? You're right about the apps, but also the hexpansion mountpoints and the settings will all be unavailable.

What are we trying to do with mount? It looks like it's designed to take over the entire FS, so wouldn't we want it to reinitialise the menu to only have the mounted apps?

b0xcat:
Hmm good point, maybe that's a better way of doing it. I just kind of quickly did this because I ran into this specific issue playing around. Will take a look

PR #301, opened 2026-05-13
Derive `app_module_name` from metadata

pikesley:

```
```

---

The reason *why* it's so broken is to do with tarballs. When we unpack a Github download we get something like:

> Note: what we get from downloading via the API is not the same thing we get from the "Download releases" webpage. Thanks, Github

`tildagon-my-app`

No owner-name, and no trailing hash.

### DIFF
--- a/modules/firmware_apps/app_store.py
+++ b/modules/firmware_apps/app_store.py
@@ -558,11 +558,6 @@ def install_app(app):
         except OSError:
             pass
 
-        # I think this is all we need
-        # So like
-        # https://whatever.forge/dave-user/tildagon-some-app
-        # resolves to
-        # dave_user_tildagon_some_app
         app_module_name = "_".join([app["id"]["owner"], app["id"]["title"]]).replace(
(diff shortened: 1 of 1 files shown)


modules/firmware_apps/app_store.py
```diff
@@ -558,7 +558,14 @@ def install_app(app):
         except OSError:
             pass
 
-        app_module_name = "_".join(prefix.split("-")[0:-1])
+        # I think this is all we need
+        # So like
+        # https://whatever.forge/dave-user/tildagon-some-app
+        # resolves to
+        # dave_user_tildagon_some_app
```
hughrawlinson: We shouldn't have this comment

hughrawlinson:
```suggestion
```

pikesley: Fixed 👍

MatthewWilkes (approved):
Tested backport, working well

PR #302, opened 2026-05-15
sim: Ensure custom import machinery doesn't hide builtin modules

### LINKED ISSUE #278: Sim: ModuleNotFoundError: No module named '_ast'
When starting the sim I get `ModuleNotFoundError: No module named '_ast'` (full stack below).

I am using pixi to run it so that I can use python 3.10 (pixi.toml below), so maybe a different environment from the devs.

A work around for me is to `import inspect` at the top of `run.py` before `sys.path` gets modified. I can make a PR for this if that's helpful.

Version: Current main f7cb5e811feee0e1cc86532e455906bd7594ed9c

**To Reproduce**
Steps to reproduce the behaviour:

1. Put below pixi.toml in the sim folder
2. `pixi run sim`

**Expected behavior**
Sim launch

**Screenshots**
```
  File "/home/sam/git/badge-2024-software/sim/run.py", line 201, in <module>
    sim_main()
  File "/home/sam/git/badge-2024-software/sim/run.py", line 197, in sim_main
    import main
  File "/home/sam/git/badge-2024-software/modules/main.py", line 3, in <module>
    import wifi
  File "/home/sam/git/badge-2024-software/modules/wifi.py", line 3, in <module>
    import asyncio
  File "/var/tmp/pixi/sim-17039356496848081362/envs/default/lib/python3.10/asyncio/__init__.py", line 8, in <module>
    from .base_events import *
  File "/var/tmp/pixi/sim-17039356496848081362/envs/default/lib/python3.10/asyncio/base_events.py", line 39, in <module>
    from . import coroutines
  File "/var/tmp/pixi/sim-17039356496848081362/envs/default/lib/python3.10/asyncio/coroutines.py", line 5, in <module>
    import inspect
  File "/var/tmp/pixi/sim-17039356496848081362/envs/default/lib/python3.10/inspect.py", line 37, in <module>
    import ast
  File "/var/tmp/pixi/sim-17039356496848081362/envs/default/lib/python3.10/ast.py", line 28, in <module>
    from _ast import *
ModuleNotFoundError: No module named '_ast'
```

**Additional context**
My pixi.toml
```
[workspace]
authors = ["Sam Tygier <samtygier@mailo.com>"]
channels = ["conda-forge"]
name = "sim"
platforms = ["linux-64"]
version = "0.1.0"

[tasks]
sim = "python run.py"

[dependencies]
python = "==3.10"

[pypi-dependencies]
wasmtime = "==39.0.0"
pygame = "==2.6.1"
requests = "==2.32.3"
```

edwinbalani:
# Description

There have been a couple of reports in the badge channel of weird failures to import Python standard library modules while trying to start the simulator, in September and again today.  I wrote off the first as a weird/broken Python installation, even though I'd noted there were some `importlib` shenanigans going on in `sim/run.py`, but the second occurrence (happening to another person) led me to look closely at the shenanigans.

I think these changes are the fix required to the sim's custom meta path finder.  However I haven't worked on one of these before and I'd appreciate some eyes and more diverse testing than just my system.

# Issue links

Resolves #278 - even though this issue was closed by the author, a `ModuleNotFoundError` trying to import `_ast` is exactly the reported issue in the second occurrence reported on the badge channel, so I'm fairly sure the cause is the same.

### DIFF
--- a/sim/run.py
+++ b/sim/run.py
@@ -43,16 +43,25 @@ class UnderscoreFinder(importlib.abc.MetaPathFinder):
         self.pathfinder = pathfinder
 
     def find_spec(self, fullname, path, target=None):
+        # Our fake 'time' module wants to import _time as a backdoor to the
+        # original time module, which is a builtin C module.  Invoke the
+        # BuiltinImporter with "time" to satisfy this.
         if fullname == "_time":
             return self.builtin.find_spec("time", path, target)
-        if fullname in ["random", "math"]:
-            return self.builtin.find_spec(fullname, path, target)
+
+        # Suppress the simulator's sys.path prefixed entries temporarily while
+        # looking for 'json' or 'tarfile'.
         if fullname in ["json", "tarfile"]:
             sys_path_saved = sys.path
             sys.path = sys_path_orig
             res = self.pathfinder.find_spec(fullname, path, target)
             sys.path = sys_path_saved
             return res
+
+        # Provide a general fallback to other builtin (C) modules.
+        # It's OK to do this indiscriminately as the BuiltinImporter won't find
+        # modules that aren't builtin.
+        return self.builtin.find_spec(fullname, path, target)
 
 
 # sys.meta_path.insert(0, Hook())


sim/run.py
```diff
@@ -43,17 +43,26 @@ def __init__(self, builtin, pathfinder):
...
+        # Our fake 'time' module wants to import _time as a backdoor to the
+        # original time module, which is a builtin C module.  Invoke the
+        # BuiltinImporter with "time" to satisfy this.
         if fullname == "_time":
             return self.builtin.find_spec("time", path, target)
-        if fullname in ["random", "math"]:
-            return self.builtin.find_spec(fullname, path, target)
+
+        # Suppress the simulator's sys.path prefixed entries temporarily while
+        # looking for 'json' or 'tarfile'.
```
edwinbalani, line 53: I couldn't find an obvious _other_ `json` or `tarfile` that would have been found if we didn't do this workaround.  However, out of the `sys.path` entries that the sim adds below, I don't have a `.../build-tilgdagon/frozen_mpy` (presumably I have to compile something).

Does anyone know if a json/tarfile in that location is what these lines are trying to avoid?

edwinbalani:
Here's the explanation I wrote in the commit message description - worth a fact-check by anyone who knows better, particularly as to the intent of the UnderscoreFinder:

```
The UnderscoreFinder is intended to override the import-time finding of
selected modules, but it is implemented by completely wrapping the
original BuiltinImporter that would be found in sys.meta_path at
startup.

Therefore the UnderscoreFinder is obliged to pass through all
find_spec() calls that it doesn't otherwise override, to avoid
accidentally breaking importation of other modules.

(It doesn't have to consult the wrapped PathFinder by default, because
we add another PathFinder to sys.meta_path.)
```

PR #304, opened 2026-05-16
Fix Neopixel addition patch

### LINKED ISSUE #303: Pattern not working?
**Describe the bug**
LED patterns are not working

Version: [for example, v1.6.0]

**To Reproduce**

1. Go to 'Settings'
2. Click on 'Next Pattern'
5. Observe error

**Expected behavior**
LED pattern shows, but instead "*name of pattern* has crashed" shows and no pattern get applied.

MatthewWilkes:
# Description

This fixes the patch to neopixel.py which was outdated, so will require `firstTime.sh` again. It also fixes a basic error in `OffPattern` where `num_leds` was set wrong.

Fixes #303

### DIFF
--- a/modules/patterns/off.py
+++ b/modules/patterns/off.py
@@ -5,4 +5,5 @@ class OffPattern(BasePattern):
     def __init__(self, num_leds=12):
         super().__init__()
         self.fps = 1
+        self.num_leds = num_leds
         self.frames = [[(0, 0, 0)] * self.num_leds]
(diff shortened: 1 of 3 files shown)


ChrisDick:
I've been having a look at fixing this one myself as I missed these when reviewing. I think there's also an off by one in patterndisplay introduced by the same pr. the offset was added to the creation of the string, but it was originally doing this by the +1 in the loop, which is still there.

MatthewWilkes:
You're right, it does! I'll update that also.

ChrisDick (approved):

PR #305, opened 2026-05-16
Derive `app_module_name` from metadata

MatthewWilkes:
Forward-porting to `main` from #301 

--------------------

```
```

---

The reason *why* it's so broken is to do with tarballs. When we unpack a Github download we get something like:

> Note: what we get from downloading via the API is not the same thing we get from the "Download releases" webpage. Thanks, Github

`tildagon-my-app`

No owner-name, and no trailing hash.

### DIFF
--- a/modules/firmware_apps/app_store.py
+++ b/modules/firmware_apps/app_store.py
@@ -558,11 +558,6 @@ def install_app(app):
         except OSError:
             pass
 
-        # I think this is all we need
-        # So like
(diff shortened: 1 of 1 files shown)


MatthewWilkes:
@hughrawlinson I've cherry-picked the commits for this as they were made against the 1.x legacy branch. Can you (or someone) approve please?

ChrisDick (approved):

hughrawlinson (approved):

PR #309, opened 2026-05-23
flasher: note the USB device name seen on an unsuccessful debug attempt

edwinbalani:
# Description

I thought I'd discovered an alternative name for the badge USB device in flasher/debug mode - it appeared in the browser as "Espressif CDC Device" - then I realised I hadn't actually held `boop` down for long enough.

It isn't too obvious that you need to power off and try again if you see this device rather than the USB/JTAG name.  Add a line to explain this.

Also, this commit wraps the HTML list items in an actual list tag.


edwinbalani:
Added some precision around when to let go of `boop` - you have to click _both_ buttons labelled "Connect" (in the esp-web-install-button, then in the Chrome permissions UI).

MatthewWilkes (approved):

PR #312, opened 2026-05-23
Spinning EMF logo on boot

pikesley:
# Display a spinning EMF Logo on Boop

I've now massively simplified this, we use just a single colour (chosen from the Official Colours), and it now seems to play fine with or without a Hexpansion plugged-in


thinkl33t:
Had a play with this:

- its very cool, great work!
- i get a very small red LED pulse towards the end, but not much else LED wise - this is on a badge with the LED brightness set to 100%
- if you boot it on a badge with one or more hexpansions plugged it, hexpansion detect and app running happens at the same moment this is trying to run, which means you only see two frames and then the main menu - we should hold the hexpansion auto-detect until after the boopscreen has finished

pikesley:
> Had a play with this:
> 
>     * its very cool, great work!
> 
>     * i get a very small red LED pulse towards the end, but not much else LED wise - this is on a badge with the LED brightness set to 100%
> 
>     * if you boot it on a badge with one or more hexpansions plugged it, hexpansion detect and app running happens at the same moment this is trying to run, which means you only see two frames and then the main menu - we should hold the hexpansion auto-detect until after the boopscreen has finished

Hey Bob

The LEDs were working correctly when I was running it as a regular app, moving it into `main` has Done Something. I will have a poke at it

pikesley:
> > Had a play with this:
> > ```
> > * its very cool, great work!
> > 
> > * i get a very small red LED pulse towards the end, but not much else LED wise - this is on a badge with the LED brightness set to 100%
> > 
> > * if you boot it on a badge with one or more hexpansions plugged it, hexpansion detect and app running happens at the same moment this is trying to run, which means you only see two frames and then the main menu - we should hold the hexpansion auto-detect until after the boopscreen has finished
> > ```
> 
> Hey Bob
> 
> The LEDs were working correctly when I was running it as a regular app, moving it into `main` has Done Something. I will have a poke at it

The LEDs now behave much better, the hexpansion thing finds me way out of my depth though (and in any case I have no hexpansions with which to test this)

pikesley:
I'm now able to test it with a real Hexpansion, and if we just reorder the things we send to the scheduler, it seems to work fine. 

@thinkl33t wanna test this if you have a moment pls?

benclifford:
The most recent commit made me uncomfortable in a "why does this change anything" sense, so I reverted it and tried on my badge with a flopagon hexpansion. I get the whole logo sequence before the flopagon app comes to the foreground.

So I'm interested to know what hexpansion (especially the firmware source) was showing the early termination problem that @thinkl33t noted above - mostly just for my understanding but I wonder if there's some weird scheduler behaviour here.

pikesley:
Maybe the behaviour is hexpansion-dependent? 

We need somebody who knows how the scheduler actually works

benclifford:
I have a medium-level understanding of how the scheduler/python async stuff works! That's why it bothers me and why I'd like to reproduce it somehow.

benclifford:
(none of this should be a blocker on the startup logo being merged and released in the wild, though...)

thinkl33t:
> So I'm interested to know what hexpansion (especially the firmware source) was showing the early termination problem that @thinkl33t noted above - mostly just for my understanding but I wonder if there's some weird scheduler behaviour here.

To be clear, the frames I was seeing were towards the middle and towards the end of the spiral, I believe the app was running but all the other async stuff running at the same time to service hexpansion startup was preventing it from rendering out.

modules/system/boopscreen/boopscreen/terminate.py
```diff
@@ -0,0 +1,7 @@
+from system.eventbus import eventbus
+from system.scheduler.events import RequestStopAppEvent
+
+
+def terminate(app):
```
MatthewWilkes, line 5: Can you put this on the app base class instead? That way it's available to all apps.

pikesley:
I will do some systematic testing later of exactly the behaviour I see -vs- the ordering of things in the scheduler (based on this one unidentified hexpansion I have)

modules/system/boopscreen/boopscreen/led_lighter.py
```diff
@@ -0,0 +1,55 @@
+from tildagonos import tildagonos
+
+from .colour_tools import rgb_from_hue
+
+
+class LEDLighter:
```
MatthewWilkes, line 4: For me, this feels like too much to have in the internal app. I think either this would be good to put in as general tools for others, but in a way that is in-line with the API of the rest of the badge (like putting a gamma correct wrapper in along with the other neopixel tools. But, that may be front-board dependent, so I am not sure. For this case, do we need it to be gamma corrected just for this logo?

benclifford:
for visual comparison, here's what I see (with the source code either as in this PR or without 510ba09) when rebooping and starting up with a flopagon installed. There's definitely a big stutter, which I assume is the flopagon app starting up and doing something that takes a while without yielding. But all the frames get rendered, just a big pause between some of them. I think that's different to what @thinkl33t is describing - because I think I'm not missing any frames, they're just rendering slower.

http://www.hawaga.org.uk/tmp/boopscreen-stutter.mp4

modules/system/boopscreen/app.py
```diff
@@ -0,0 +1,78 @@
...
+        if self.logo.faded():
+            eventbus.emit(PatternEnable())
+            terminate(self)
+
+    def draw(self, ctx):
+        """Draw."""
+        ctx.rotate(self.rotation)
+        self.overlays = []
+        self.overlays.append(
+            Background(colour=(0, 0, 0), opacity=conf["background-opacity"])
```
MatthewWilkes: This feels excessive, we have a built-in clear_bg function, I believe, and adding it to an overlay adds a lot of indirection. We could just clear the bg here and drop the whole background class.

modules/system/boopscreen/app.py
```diff
@@ -0,0 +1,78 @@
...
+        )
+
+        self.overlays.append(self.logo)
+
+        self.draw_overlays(ctx)
+
+    def scan_buttons(self):
```
MatthewWilkes: This is fine, but you could also just listen to the event and cancel the app, to simplify the work on update?

modules/system/boopscreen/app.py
```diff
@@ -0,0 +1,78 @@
...
+        if self.fading:
+            self.logo.fade()
+
+        elif self.leds.brightness < conf["leds"]["max"]:
+            self.leds.brightness += conf["leds"]["increment"]
+
+        self.leds.from_rgb([int(x * 255) for x in colour])
+
+        if self.logo.faded():
+            eventbus.emit(PatternEnable())
```
MatthewWilkes, line 55: This will only fire if the app is not cancelled, right?

modules/main.py
```diff
@@ -9,16 +9,19 @@
 from system.launcher.app import Launcher
 from system.power.handler import PowerEventHandler
 from system.power.app import PowerManager
+from system.boopscreen.app import BoopSpinner
 
 from frontboards.twentyfour import TwentyTwentyFour
 
+# Start expansion interface
+scheduler.start_app(HexpansionManagerApp())
```
MatthewWilkes, line 37: I'm not sure that moving this earlier makes sense. I'd be tempted to have it later (as it was) and maybe put a short delay on. That would also help with hexpansions loading causing the menu to jitter.

modules/main.py
```diff
@@ -9,16 +9,19 @@
...
+scheduler.start_app(BoopSpinner(), always_on_top=True)
 
 # Start front-board interface
 scheduler.start_app(TwentyTwentyFour())
 
-# Start expansion interface
-scheduler.start_app(HexpansionManagerApp())
-
 # Start led pattern displayer app
 scheduler.start_app(PatternDisplay())
```
MatthewWilkes, line 43: If this is starting after the boopspinner, then could the patterndisable be missed?

benclifford:
which makes me wonder is the Mystery Hexpansion firmware doing something that would interfere with SPI or with ctx or trying to dominate the badge in some weird way at firmware app startup.

modules/system/boopscreen/boopscreen/terminate.py
pikesley: Sure 👍

modules/system/boopscreen/boopscreen/led_lighter.py
pikesley: Almost certainly not, all my apps are *wlidly* overengineered

benclifford: i have wanted gamma correct elsewhere so I would +1 making more generally available. wrt different board types, I don't have a really strong feel for how much different neopixel hardware needs different gamma corrections but I've never noticed any really big difference across my handful of projects.

modules/system/boopscreen/app.py
pikesley: Sounds good 👍

modules/system/boopscreen/app.py
pikesley: Um, correct. I think

modules/main.py
pikesley: Delay within the `spinner` app?

modules/system/boopscreen/app.py
pikesley: 👍

thinkl33t:
> which makes me wonder is the Mystery Hexpansion firmware doing something that would interfere with SPI or with ctx or trying to dominate the badge in some weird way at firmware app startup.

Nothing particularly hardcore: https://github.com/emfcamp/hexpansion-firmwares/blob/main/0x1337/0xaeae/app.py

modules/system/boopscreen/boopscreen/led_lighter.py
kliment: We don't really need it for the neopixels but the 2026 frontboard has them shining through yellowish PCB material and that can definitely do with some correction.

pikesley:
I will apply all of these optimisations this evening and we will see where we get to

modules/main.py
benclifford: in my understanding of the concurrency here: (not sure if i have it all in my head properly)

everyone's init method runs at the time of these start_app calls, so i) a callback are registered by PatternDisplay and ii) an event is sent by BoopSpinner and put in the queue.

but no event dispatch actually happens until later on - at the `run_forever` call at the end. (more generally *no* code marked `async` runs until then, I think right?)

in which case, it doesn't matter which order i) and ii) happen because the pairing of the event and callback happens at the run_forever stage?

pikesley: That's what I thought too *but it definitely behaved differently with different ordering*. Consistently enough that I wasn't just getting lucky with the Concurrency Dice

benclifford: are you saying that you also saw problems with patterndisable ordering? (not the screen update stuff)

modules/system/boopscreen/boopscreen/terminate.py
pikesley: I will do this in a separate PR

modules/system/boopscreen/boopscreen/led_lighter.py
pikesley: Gamma correction removed for now

modules/system/boopscreen/app.py
pikesley: I have nuked this and moved to the native `clear_background` (which I guess will pick up the 2026 background-colour when the time comes)

modules/system/boopscreen/app.py
pikesley: This is gone now

pikesley:
Just rebased this onto `main` and some behaviour has slightly changed. Will have a look

benclifford:
I'm actively running this PR as part of my daily firmware bodgestack - let me know if you want me to comment on how i see it behaving.

pikesley:
> I'm actively running this PR as part of my daily firmware bodgestack - let me know if you want me to comment on how i see it behaving.

pls :+1:

PR #314, opened 2026-05-27
add retry after 100ms on EEPROM scan

sammachin:
Adds a retry to the EEPROM scan if not found first time, with a 100ms delay.

I've implemented the EEPROM in my RP2040 microcontroller on my hexpansion but this takes a good 10ms to be ready after power is applied so the previous scan was sometimes too slow to find it and launch the app.

### DIFF
--- a/modules/system/hexpansion/app.py
+++ b/modules/system/hexpansion/app.py
@@ -29,6 +29,7 @@ from egpio import ePin
 from system.eventbus import eventbus
 from machine import I2C
 from events.input import Buttons
+import asyncio
 import vfs
 import sys
 import settings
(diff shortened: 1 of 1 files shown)


Jonty:
We could just always wait 100ms as it's barely even perceptible, and I bet it would increase reliability for other weird hexpansions too!

sammachin:
Yeah the 2.0.0-alpha-3 already seems better at first time detect anyway as there's some other stuff that goes on now for detection that I think helped the race condition.
Retry feels neater though and it will only execute this path if the first scan fails, so if it now has twice the chances to detect

MatthewWilkes (approved):
Great, nice minimal change.

PR #315, opened 2026-05-27
Frontboard2026

ChrisDick:
# Description

new frontboard code, including:
* detection of attached frontboard
* auxiliary port expander
* compass added to imu
* touch sensor driver
* touch feedback on the leds
* app to raise events

app currently polls the touch ic, but has the callback setup in c for the interrupt driven version.

compass not tested yet.


modules/events/input.py
```diff
@@ -56,6 +56,25 @@ def find_parent_in_group(self, group):
     "DOWN": Button("DOWN", "System"),
     "LEFT": Button("LEFT", "System"),
     "RIGHT": Button("RIGHT", "System"),
+    "A": Button("A", "System"),
```
benclifford: is the button rearrangement that this is part of going to do weird stuff to the app store code input dialog, which uses these buttons that I think are being renamed?

ChrisDick: possibly, one of the reasons for putting the PR up is to get feedback on this stuff. it's really a draft PR, but if I mark it as draft, the CI doesn't run and I need others with the new top board to test things like the touch response on multiple badges for tolerance checks and to test new boards as they're made. I still need to debug the compass, do some docs and get it all reviewed, so there's time for feedback.

ChrisDick:
I've reverted BUTTON_TYPES to the 6 system types and remapped system button types to both buttons and joystick. Named buttons the same as 2024 names and added new names for the joystick. This lets existing core and user apps stay unchanged and have both button or joystick control. Buttons for the 2024 frontboard are already named A-F, anyone wanting to have different uses for the buttons and joystick can do. usage could be:
generic control
```python
```
or 
```python
if "FIRE" in event.button.name:
```

ChrisDick:
This is now ready for review.

b0xcat:
Just tested this with the rev1 2026 frontboard and it works great! 

The only thing I noticed was that when an app takes over the LEDs it does not go back to normal after and the touch indicators stop working. 

Will take some time to go through the code as soon as I can!

MatthewWilkes:
As discussed in the last badge call, there are still things to do on this, especially working out LED highlight behaviour, and maybe some tweaking of button hierarchies, but it's time to merge it.

Tested and working on multiple devices for me, we should get it into people's hands.

MatthewWilkes (approved):
Approved, as above

PR #316, opened 2026-05-27
Add brand font and additional glyphs

walkerdanny:
# Description

Added the brand font as per #23 but with a custom Raleway build including additional glyphs for `(⬡,⬢,▲,▶,▼,◀) ` 

This allows inline rendering of hexagons and direction triangles alongside text, which is cool and useful for UI stuff.

Since there are new glyphs, the font has been renamed to EMF Camp Font to comply with the OFL, so I don't think there's any issues there.

I've flashed this onto my badge and it works fine, but I have since broken my build toolchain so it could maybe use a tester. Toolchain for editing and baking the font is working though so more glyphs can be added if desired.

An image showing the working extra glyphs on the main badge menu

### DIFF
--- a/components/ctx/ctx_config.h
+++ b/components/ctx/ctx_config.h
@@ -97,9 +97,10 @@
 
 #define CTX_MAX_FONTS 3
 
-#include "Arimo-Regular.h"
-#define CTX_FONT_0 CTX_STATIC_FONT(Arimo_Regular)
-
+#include "EMFCampFont.h"
+#define CTX_FONT_0 CTX_STATIC_FONT(EMF_Camp_Font)
+// #include "MaterialIcons.h"
+// #define CTX_FONT_1 CTX_STATIC_FONT(MaterialIcons)
 // there is room for a mono font, but then the partition is full
 //
 //#include "Comic-Mono.h"
--- a/components/ctx/fonts/EMFCampFont.h
+++ b/components/ctx/fonts/EMFCampFont.h
@@ -1,0 +1,12487 @@
+#ifndef CTX_FONT_EMF_Camp_Font
+/* glyph index: 
+ !"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghi
+  jklmnopqrstuvwxyz{|}~ ¡¢£¤¥¦§¨©ª«¬­®¯°±²³´µ¶·¸¹º»¼½¾¿ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔ
+  ÕÖ×ØÙÚÛÜÝÞßàáâãäåæçèéêëìíîïðñòóôõö÷øùúûüýþÿǩ←↑→↓↔↕↖↗↘↙⇩⏎␈␏□▲△▶▼◀◇○♣✕⬡⬢⬣⭍䇩燩
+  臩  */
+  
+
+/* Copyright 2024 tildagon Project Authors.
+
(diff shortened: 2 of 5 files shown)


MatthewWilkes:
I will look at this properly when I get a chance, but just from the quick look on my phone, has this changed the micropython submodule pin? It looks like it may have?

theorbtwo:
I think it'd be more useful to have less-ambiguous arrows, rather than equilateral triangles, and to have more of them, one for each of the major buttons and connectors/flats?

walkerdanny:
> I will look at this properly when I get a chance, but just from the quick look on my phone, has this changed the micropython submodule pin? It looks like it may have?

You're right, I spent a while fighting this repo last night and it looks like I failed. I'll have another look. Marking this as draft for now.


> I think it'd be more useful to have less-ambiguous arrows, rather than equilateral triangles, and to have more of them, one for each of the major buttons and connectors/flats?

This would be getting into custom unicode territory I think, [unicode arrows ](https://unicode.org/charts/nameslist/n_2190.html) only seem to include 45 degree arrows, where all corners and flats would need 30 degree increments.

walkerdanny:
Micropython version should be fixed now 🤞

walkerdanny:
I've added in the spider emoji to the font, but either ctx or micropython doesn't like rendering characters in a unicode block that high (I think because it requires an extra byte).

Two proposed solutions:
- Use the private use area and put the spider somewhere in the range `0xE000 - 0xF8FF`, but this won't render on computers when debugging (or will be tofu)
- We misuse the unicode blocks slightly to put the spider at `0x81E9` (8LEG) which is easier to remember. This means we'd be squatting on a space intended for a chinese character 臩, which is a very rare old character, for which the best definition I can find is "To run away in fright".

I've tested the second solution and it works 🕷🕷🕷:
spiderhanzi

thinkl33t:
> We misuse the unicode blocks slightly to put the spider at 0x81E9 (8LEG) which is easier to remember. This means we'd be squatting on a space intended for a chinese character 臩, which is a very rare old character, for which the best definition I can find is "To run away in fright".

This is my favourite PR ever

Jonty:
This is amazing.

If possible could we squash all these commits before merging? It's a lot of dev noise at the moment.

walkerdanny:
Sure, I'll squash it for the final version. Re-marking as draft until then.

walkerdanny:
I think I'm close to a final character set. Now included:

- Full set of pointing arrows `←↑→↓↔↕↖↗↘↙`
- Hexagons including a horizontal one which renders as a hexpansion `⬡⬢⬣`
- Pointing triangles if that's your thing `▲▶▼◀`
- All non-text symbols featured on the official keyboard hexpansion `⏎␈␏□△◇○♣✕⭍` (the last one is where the Solder Party logo lives)
- Spider (`0x81E9`), duck (`0x21E9`) and blåhaj (`0x01E9`)
- Exactly two bonus glyphs with other quantities of legs
- Space for EMF logo when the 2026 version is available

I've mapped as many glyphs as possible onto the closest unicode equivalents so they'll be renderable as _something_ when developing.

Testing welcomed, all builds and looks ok on my badge. Haven't tested the sim as I don't have a working setup for it.

walkerdanny:
Added the 2026 logo now it's public and added licenses for the Solder Party keyboard icons. Should be ready to merge.

finalfont

andypiper:
🦆 this is superb.

thinkl33t:
Looking at merging this shortly ready for the next alpha release - any chance you could do a rebase and tidy up the commits a little?

(I'd also be happy enough with one squashed commit)

thinkl33t (approved):

PR #319, opened 2026-05-30
Remove circular dependency in hexpansion code

thinkl33t:
There was a circular dependency between system.hexpansion.app and  system.hexpansion.util, which made the user unable to import  system.hexpansion.util, and stopped the hexpansion development tools  from working properly.

This should solve the import issue, and at least in my testing the  tooling works again.


MatthewWilkes (approved):
Makes sense, yes. Thanks!

PR #320, opened 2026-05-31
WASM badge firmware build

Jonty:
This adds support for building micropython as WASM, along with native browser CTX support and all of our python firmware baked in. It does _not_ swap in stubs for python code that expects to be on native hardware to ensure that it's accurate to the main firmware, and that any environment using it can override with suitable stubs for where it's being run.

There are three patches to upstream micropython applied as part of this:
* A patch to the micropython JS bindings to fix an async bug. The fix is already proposed upstream by someone else but not yet merged.
* The webassembly time implementation so that it matches the ESP32 in having the epoch set to 2000 rather than 1970.
* The webassembly "sleep" implementation to use the emscripten version, so we actually yield back to JS.

This also removes two modules from the firmware that we are not using and would have required annoying workarounds:

* st3m_scope support (a broken oscilliscope from the flow3r firmware)
* gc9a01py, a pure python screen driver that was replaced with ctx

See this for something that uses it: https://github.com/emfcamp/tildagon_web_emulator/commit/78b94b3cb37ccee9b04cb4ce3bdc04bf446293fd


patches/micropython_wasm.diff
```diff
@@ -0,0 +1,47 @@
+diff --git a/ports/webassembly/mphalport.c b/ports/webassembly/mphalport.c
```
MatthewWilkes: @Jonty Is there a specific reason to keep this patch split from the other micropython one? Just thinking that if they are in the same file, they are easier to keep up to date

Jonty: Now you say that ...no not really. I was trying to split it up by "things that might get fixed upstream soon" but that's stupid isn't it.

MatthewWilkes: I think it will be easier as one, but I understand the instinct. Otherwise LGTM

Jonty: Merged into one file!

MatthewWilkes (approved):

Issue #317, opened 2026-05-30
Add & document a way for hexpansions to add things to the main menu.


thinkl33t:
**Is your feature request related to a problem? Please describe.**

Some hexpansions have a foreground-capable app, for example to change configuration settings or read received messages.  Currently there isn't a good way to launch / relaunch these apps without replugging the hexpansion.

**Describe the solution you'd like**

A documented method that apps can use to add relative links to apps running on themselves to the main menu.  Ideally there should be some magic to pull the current eeprom mountpoint as part of this.

hairymnstr:
The Floppagon is a perfect example use case.  The default app that is meant to live in the EEPROM offers 3 options:
* Mount FLASH storage
* Unmount FLASH storage
* Format FLASH storage
A typical use case would be to run the app on insert and use the "Mount FLASH" option, then use something on the Flash (maybe the Hexfile app) then when you're done go back to the main menu and "Unmount FLASH storage" to safely remove it.  Having a way to register it in the main launcher would facilitate this work flow.

walkerdanny:
See also #184

sammachin:
My scenario is that my app.py is basically a client to my hexpansion communicating with it over uart.

So having the app launch when the hexpansion is plugged in is great but users also need to be able to get back to the client to interact with the hexp again from the menu, ideally withough having to replug the hardware as this restarts the code running on the hexp and some features like time are lost, as well as being fiddly and prone to wearing parts.

Ideally the app should appear to users in the main menu exactly as if it had been installed from the app store, shipping the app on the eeprom just allows me to keep the client and hexpansion firmware features in sync without needing to version things. I can add features to my firmware and the coresponding ui in the app all in one release.

Adding a manifest.json to the littlefs filesystem on the eeprom would be one approach that could work for me, so if thats present as well as the app.py it adds to the menu?
(I've got a full 64KB of EEPROM fs space to play with in my implementation so this might not work for all)

Issue #184, opened 2024-07-04
Add apps loaded from Hexpansion EEPROM to the launcher menu


walkerdanny:
I've developed a couple of apps now that can load from a hexpansion EEPROM, and they auto-run on insertion.

If the app displays a UI, it can foreground itself and the user can interact with it. If the app is then minimised, there is no way of returning to the app even though it is still running.

Ideally when an app is launched from a hexpansion EEPROM it should be added to the launcher menu along with the other installed apps (maybe with an identifier to show it's a hexpansion app, or within a "Hexpansions" submenu) so that the user can re-open it when minimised.

Maybe this could be done by exposing part of the Launcher API to the app, so the app can add itself into the menu. That way hexpansions with no UI or that only require a background service won't get added.

walkerdanny:
I've added a first version of this functionality in [this commit](https://github.com/walkerdanny/badge-2024-software/commit/b3061e149ca081b43d5022f3828fdcfe0f765507) which involved adding in 4 more hexpansion events:

- `HexpansionAppLauncherAddEvent`
- `HexpansionAppLauncherRemoveEvent`
- `HexpansionAppRequestStartEvent`
- `HexpansionAppRequestStopEvent`

 `HexpansionAppLauncherAddEvent` and `HexpansionAppLauncherRemoveEvent` can be emitted by an app which is loaded from a hexpansion EEPROM to add themselves to the launcher menu. The launcher functionality is extended to allow it to foreground these apps once they have been minimised.

`HexpansionAppRequestStartEvent` and `HexpansionAppRequestStopEvent` were added so that the launcher can interact with hexpansion apps properly, because they're usually launched from `HexpansionManager` rather than the launcher itself, so these events allow app start and stop events to be passed between the two.

I could do with a hand testing this as I'm getting some weird bugs that are hard to pin down - sometimes i get `Foreground request ignored for app that's not running: <TestApp object at 3c2338b0>` which seems like a race condition somewhere that needs to be handled, and other times the app doesn't get removed from the launcher properly when the hexpansion is physically removed.

This branch also implements some additional helper functions for #140 so apps can import HexpansionManager and see which other slots are populated etc. This might not be the best way of doing this but it was easy to include.

A minimal example of a hexpansion EEPROM app for testing purposes is:
```python
import app
from system.eventbus import eventbus
from events.input import BUTTON_TYPES, ButtonDownEvent
[...]

PR #321, opened 2026-06-01
Allow hexpansions to add themselves to launcher

walkerdanny:
# Description

As per #317 and #184 this is a working (as far as my testing has revealed) way for hexpansion EEPROM apps to add themselves to the launcher.

It adds four new event types to add a hexpansion to launcher, remove a hexpansion from launcher, start a hexpansion app and stop a hexpansion app. Not crazy about the names of these so please suggest better ones if you have ideas, naming things is hard.

This is done on a port number basis, so multiple hexpansions of the same type should each be able to have their own instance of the app. That felt like the right approach.

The hexpansion glyph allocated in #316 is prepended to app names so they appear in the launcher with a hexpansion next to them (eventually)

I get the occasional glitch with this that sorts itself out after a while, but I think this happens when a hexpansion isn't inserted properly or gets inserted and removed quickly, so it's probably not unique behaviour to this branch.

A minimum working hexpansion app for testing is:

```python
import app
from system.hexpansion.events import HexpansionAppLauncherAddEvent
from app_components import clear_background
from system.scheduler.events import RequestForegroundPushEvent

class TestApp(app.App):
    def __init__(self, config=None):
        self.app = app
        self.hexpansion_config = config
        self.foregrounded = False
        self.add_to_launcher()

    def add_to_launcher(self):
        # This is the only special thing your hexpansion app needs to do besides importing the event:
        eventbus.emit(HexpansionAppLauncherAddEvent(self.hexpansion_config.port, "Hexpansion Test App"))

    def update(self, delta):
        if not self.foregrounded:
            print("Requesting foreground push")
            eventbus.emit(RequestForegroundPushEvent(self))
            self.foregrounded = True
            self.button_states.clear()
            self.minimise()

    def background_update(self, delta):
        # Do nothing in the background
        pass

    def draw(self, ctx):
        clear_background(ctx)
        ctx.font_size = 20
        ctx.text_align = ctx.CENTER
        ctx.rgb(255,255,255).move_to(0, 0).text(f"Hexpansion {self.hexpansion_config.port} app")
        return None

__app_export__ = TestApp
```


modules/system/hexpansion/app.py
```diff
@@ -38,7 +41,7 @@
 active_back_leds = [False] * 6
 
 
-def Hexspansion_inserted(epin):
```
MatthewWilkes, line 36: lol

modules/system/launcher/app.py
```diff
@@ -80,13 +85,19 @@ def list_user_apps():
                 apps.append(app)
         return apps
 
+def list_hexpansion_apps(): # List apps loaded from hexpansion EEPROMS
+    #print (scheduler.apps[1].hexpansion_apps)
```
MatthewWilkes: @walkerdanny Is this meant to be commented out?

walkerdanny: Hmm I think this was just in here for debugging since it looks like a bit of a fudge

walkerdanny:
Have tested some more and it wasn't working properly when several of the same hexpansion were plugged in and all add the same app to the launcher. When bringing the app back to the foreground from the launcher, the launcher foregrounds the first instance of the app it finds because of:
https://github.com/emfcamp/badge-2024-software/blob/2e404586e3cc25d97b4a56b484423a759950b189/modules/system/scheduler/__init__.py#L138-L142

The problem here is that given only a list of `App` objects we can't tie any individual app to a hexpansion. Currently the `App` class doesn't store the `HexpansionConfig` object that gets passed to it at first launch, it's up to the user to do that and the variable name they choose could be anything (or they might not do it at all). I've added this into the base `App` class so that it's consistently stored at `app.hexpansion_config` and defaults to `None`. This will require hexpansion app writers to do `super().__init__(config)` if they want to add to launcher, but this can be documented. I _think_ it shouldn't break anything existing with existing hexpansion apps but this will need a lot of testing.

The `RequestForegroundPushEvent` now also takes an optional `port` argument so that the scheduler knows which version of an app to foreground if multiple of the same app are running. I've also prepended the slot number to the name that appears in the launcher to better distinguish them.

MatthewWilkes:
I am a bit confused, as it uses is for the comparison, the foreground push should distinguish between the instances. Is that not what you are observing?

walkerdanny:
I thought the same but I could only ever get it to launch the first result in the list of running apps. It's very possible that the bug was elsewhere, but the latest changes definitely fix it. Could it be that the `index` method can't distinguish between instances?

MatthewWilkes:
I think I need to have a play, I feel like it should be possible for apps not to need this, and that there must be one or more bugs that we are working around rather than fixing. It feels wrong that the scheduler is mixing up apps, and if it were not doing that then it would all work smoothly.

walkerdanny:
I think you're right, I'll do a bit more digging with fresh eyes tomorrow:
```
>>> scheduler.apps
[<TwentyTwentyFour object at 3c24c500>, <HexpansionManagerApp object at 3c24c5a0>, <PatternDisplay object at 3c243130>, <Launcher object at 3c2434e0>, <NotificationService object at 3c260f80>, <PowerManager object at 3c245a60>, <TestApp object at 3c255390>, <TestApp object at 3c25e6b0>]
>>> scheduler.apps[6]
<TestApp object at 3c255390>
>>> scheduler.apps[7]
<TestApp object at 3c25e6b0>
>>> scheduler.apps.index(scheduler.apps[7])
7
>>> scheduler.apps.index(scheduler.apps[6])
6
```

walkerdanny:
Have traced this through, the problem is actually in adding apps with identical names to the launcher causes the first instance to always be launched because they're selected by name:
https://github.com/emfcamp/badge-2024-software/blob/2e404586e3cc25d97b4a56b484423a759950b189/modules/system/launcher/app.py#L173-L177

The way that my mods fixed this in passing was by adding the hexpansion slot number to the name so the launcher can distinguish between instances of the same app. The easiest thing to do is keep that feature and bin the other changes I made.

MatthewWilkes (approved):

MatthewWilkes:
I've tested this and it's working well, so going to merge. Thanks @walkerdanny

PR #323, opened 2026-06-02
patch NimBLE for bonding

sammachin:
This is a patch to the NimBLE stack in micropython, my app  needs an encrypted, paired BLE connection, and on the 2.0.0-alpha3 firmware the badge hard-resets the instant `gap_pair()` is called.
The crash is a MicroPython asserting:
```
assert failed: ble_secret_store_read modbluetooth_nimble.c:1948
    (ble_addr_cmp(&key->sec.peer_addr, BLE_ADDR_ANY))
```
In `extmod/nimble/modbluetooth_nimble.c`, 

the bond store read handler handles `BLE_STORE_OBJ_TYPE_PEER_SEC` lookups with `peer_addr == BLE_ADDR_ANY` 
by iterating records by index, but  for `BLE_STORE_OBJ_TYPE_OUR_SEC` it **asserts** that an address must be set.

During pairing NimBLE enumerates "our security" records by index (with `BLE_ADDR_ANY`) to count existing bonds before storing a new one. 
With bonding enabled - and especially with an iPhone using a Resolvable Private Address that by-index 
`OUR_SEC` read happens and trips the assert, which (with the badge building assertions to `abort()`) is an immediate reset. 
MicroPython's own CI pairs two ESP32s with fixed addresses and never exercises this path, which is why the bug shipped.

I'm going to submit an upstream PR to micropython to fix this as well but I'm not sure it would get merged there in time for EMF and I'l really like to ship this app for the event.

Demo of the app (running on a patched badge) https://www.youtube.com/watch?v=791P1Sn_h2k


scripts/firstTime.sh
```diff
@@ -4,6 +4,7 @@ set -e
 pushd micropython
 git reset --hard
 git apply ../patches/micropython.diff
+git apply ../patches/nimble-our-sec-by-index.diff
```
MatthewWilkes: @sammachin Would you please add this to micropython.diff rather than adding a new patch file? That way, when we are managing changes we can do ` git diff > ../patches/micropython.diff` to keep them synced when we are writing patches? Otherwise, looks sensible to me.

MatthewWilkes (changes requested):

scripts/firstTime.sh
sammachin: @MatthewWilkes done

MatthewWilkes (approved):
Approved, now the patches are merged

PR #325, opened 2026-06-02
Wrap notifications based on words if we can

Jonty:
Because it rea
lly sucks when they are wrappe
d in the middle of a word

### DIFF
--- a/modules/app_components/notification.py
+++ b/modules/app_components/notification.py
@@ -44,18 +44,25 @@ class Notification:
 
     def get_text_for_line(self, ctx, text, line):
         width_for_line = 240
-        if line < (len(self.width_limits)):
+        if line < len(self.width_limits):
             width_for_line = self.width_limits[line]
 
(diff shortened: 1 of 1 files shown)


MatthewWilkes:
@Jonty Testing this, the text `"Updatededed testing"` wraps onto three lines, as:
```
    Updatedede
    d
    testing
```

Jonty:
> @Jonty Testing this, the text `"Updatededed testing"` wraps onto three lines, as:
> 
> ```
>     Updatedede
>     d
>     testing
> ```

This'll learn me for totally rewriting it because I was unhappy with the first implementation and then not actually testing it because I was sure it was right.

MatthewWilkes:
It absolutely will not learn you. None of us ever truly learn that.

Jonty:
@MatthewWilkes I think that'll do it.

MatthewWilkes (approved):
This is satisfying.

PR #327, opened 2026-06-02
Make App Store codeinstall see keyboard buttons

benclifford:
Before this PR, the code input dialog only did directional frontboard A-F buttons, even though codes are often represented as numbers.

It did not respond to button pushes for actual number buttons on tildagons with those additional button types
(in my case, my usb serial keyboard driver, but I guess also any other external keyboard? I'm not sure? it's also "broken" behaviour inside the web emulator which has a keyboard driver)

After this PR, you can use the frontboard A-F buttons as before or press numerical keys if you have them.

### DIFF
--- a/modules/firmware_apps/app_store.py
+++ b/modules/firmware_apps/app_store.py
@@ -488,6 +488,8 @@ class CodeInstall:
         eventbus.on(ButtonDownEvent, self._handle_buttondown, app)
 
     def _handle_buttondown(self, event: ButtonDownEvent):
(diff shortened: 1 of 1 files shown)


MatthewWilkes:
I think the codes might be switching to A-F, otherwise this looks great.

MatthewWilkes:
Actually, let's merge it, then whoever implements that can just change the string.

MatthewWilkes (approved):

Issue #318, opened 2026-05-30
Add & Document a way to run a teardown script when a hexpansion is unplugged


thinkl33t:
**Is your feature request related to a problem? Please describe.**

When a hexpansion is unplugged which has changed something about the badge's config, this may need undoing.

**Describe the solution you'd like**
An `_teardown` method or similar that, if it exists is run when a hexpansion is unplugged would be valuable

**Additional context**

for example, the flopagon installs its spi eeprom (equivilent code path):

``` 
        self.hspi = SPI(1, 10000000, sck=self.config.pin[1], mosi=self.config.pin[2], miso=self.config.pin[3])
        self.flash = FLASH(self.hspi, self.cspins, cmd5=False)
        os.mount(self.flash, MOUNTPATH)
```

It would be useful, if the flopagon is unplugged, to be able to automate an `os.umount()` and the like

mbooth101:
Is it not sufficient to do this in the HexpansionUnmountedEvent or HexpansionRemovalEvent handler?

hairymnstr:
I've been digging into this a bit.  There's no clean way to do this with the current events system from an app that is launched from the Hexpansion.

The sequence of events when a hexpansion is removed goes something like:
1. Detect pin IRQ triggered which generates the HexpansionRemovedEvent
2. Event handlers for that event are called, if your app has an event handler it *might* get called first, but it might not because the system has a handler
3. The system handler will emit a RequestStopApp event for your hexpansion app, but trapping that seems impossible because it doesn't wait and then
4. immediately deletes the app and its resources, so even if the app did have a handler for RequestStopApp it would have been deleted
5. Further cleanup of imports and mount points goes on until finally
6. HexpansionUnmountedEvent is emitted.  By this point your hexpansion app is long gone and any handler it had for that event is deleted.

I think we need to have a direct call to a tear-down handler instead of emitting the RequestStopAppEvent here: https://github.com/emfcamp/badge-2024-software/blob/b661be4dfb6e27b8089077c228fd99aa99c8f085/modules/system/hexpansion/app.py#L196

hairymnstr:
Of course, this mysterious function that should be called when an object is deleted is called a destructor, and is part of all OO languages. 🤦 

The solution should just be to implement `__del__` in the hexpansion app class.  I will try it out on the Flopagon and assuming it works as expected I'll add some demo code to the docs.

hairymnstr:
[...]

PR #328, opened 2026-06-05
Add call to deinit hexpansion apps

hairymnstr:
# Description

Adds a call to optional method `deinit` when a hexpansion is removed.  This can be used for tear-down things like freeing hardware peripherals or unmounting extra file systems.  It is more reliable to have an explicit deinit function call because the python destructor method `__del__` may or may not be called depending on when the garbage collector runs.

Addresses issue #318


MatthewWilkes (approved):
Thanks!

PR #330, opened 2026-06-09
Add configurable menu highlight colour

Coedy:
If you install multiple long named apps from the store its hard to tell which one you are highlighting as there isnt much room for the app name to enlarge.
Focused menu items render in a user-selectable highlight colour  (yellow by default) instead of plain white, making it much easier  to see which item is selected, especially for long app names.

Works on main menu, sponsors, & any level within the appstore as uses the deault methods for navigating through lists.

image

- Six colours available: Yellow, Pink, White, Green, Orange, Blue.
- Setting _should_ persist reboots/reloads of the Sim via the existing settings system and can be changed in the Settings app under 'Menu highlight' section.
  
Also fixes a simulator bug where bare root paths like /settings.json were not redirected to /tmp/sim/, preventing settings from persisting across simulator sessions and crashing the sim when trying to exit after a change (Permissions Error).

Only tested on the Sim. Created using Claude hence the bad commenting, as I had tried this twice previously and wasn't able to get it working.
This is my first 'real' PR so apologies if I've missed any guidelines — happy for any feedback.

### DIFF
--- a/modules/firmware_apps/app_store.py
+++ b/modules/firmware_apps/app_store.py
@@ -22,6 +22,7 @@ from requests import get
 from system.eventbus import eventbus
 from system.launcher.app import (
     APP_DIR,
+    APP_INSTALL_DIR,
     load_info,
     InstallNotificationEvent,
 )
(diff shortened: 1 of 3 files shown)


Coedy:
I've stripped back the comments to match the style of the rest of the repo sorry.

MatthewWilkes:
@Coedy Thanks for this, however, I'm going to hold off on it for now, as we're working on merging in the 2026 functionality, and part of that includes a themed highlight colour for the menu. As such, some of this would need tweaking.

Coedy:
No problem! Thanks for the update! Looking forward to the new badge front and software 😄

ChrisDick:
#344 has been merged now if you want to take another look at this?

PR #331, opened 2026-06-09
Missing await on async sleep calls

Jonty:
Spotted an async sleep missing an await (so it wouldn't do anything), grepping found another missing await, and one that shouldn't be async at all.

### DIFF
--- a/modules/frontboards/twentyfour.py
+++ b/modules/frontboards/twentyfour.py
@@ -36,7 +36,7 @@ def buttondown(epin):
                 now = time.ticks_ms()
                 if TwentyTwentyFour.hexpansion_states[hexindex] is None:
                     TwentyTwentyFour.hexpansion_states[hexindex] = now
-                    eventbus.emit_async(HexpansionInsertionEvent(port=hexindex))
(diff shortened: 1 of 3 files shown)


modules/frontboards/twentyfour.py
```diff
@@ -36,7 +36,7 @@ def buttondown(epin):
                 now = time.ticks_ms()
                 if TwentyTwentyFour.hexpansion_states[hexindex] is None:
                     TwentyTwentyFour.hexpansion_states[hexindex] = now
-                    eventbus.emit_async(HexpansionInsertionEvent(port=hexindex))
+                    await eventbus.emit_async(HexpansionInsertionEvent(port=hexindex))
```
ChrisDick: this await breaks the buttons, should be non async emit instead. this is one of the callbacks used by the c code, so just goes on the mp scheduler.

ChrisDick (changes requested):

modules/frontboards/twentyfour.py
Jonty: Good catch! I didn't think to check if it even _should_ be async...

ChrisDick (approved):

PR #333, opened 2026-06-13
Fixed crash on wifi setting reset if wifi_wpa2ent_username not set

walkerdanny:
# Description

Hitting "Reset WiFi" in the settings app causes a crash to REPL with `KeyError: wifi_wpa2ent_username` if the badge has previously been connected to a non wpa2 enterprise network, becaue `wifi_wpa2ent_username = None`. Added a check for this when resetting which prevents this.

### DIFF
--- a/modules/settings.py
+++ b/modules/settings.py
@@ -26,7 +26,8 @@ def set(k, v):
     if _settings is None:
         load()
     if v is None:
-        del _settings[k]
+        if k in _settings:
+            del _settings[k]
     else:
         _settings[k] = v
     _modified = True


walkerdanny:
After more than 30 seconds of thinking: should this button reset to the camp wifi details? That's probably more useful than setting everything to None and means when people who got their badges in 2024 return in 2026 they can get on the wifi easily.

MatthewWilkes:
@walkerdanny The behaviour when it is unset is to use the camp details, so it would do this. I'd be inclined to say that the bug here is in settings.py rather than the app, though. The bug happens in the underlying set(), shouldn't that be fixed so you don't have to be aware of the current state to set a value of None?

walkerdanny:
> @walkerdanny The behaviour when it is unset is to use the camp details, so it would do this. I'd be inclined to say that the bug here is in settings.py rather than the app, though. The bug happens in the underlying set(), shouldn't that be fixed so you don't have to be aware of the current state to set a value of None?

Yep good point. I'll set my critical thinking timeout to 60 seconds while I'm at it.

walkerdanny:
Now fixed in a nicer way

thinkl33t (approved):
Brill - this should fix the problem nicely.

PR #341, opened 2026-06-19
Fix step counter & add reset function

sammachin:
The imu. step_counter_read() always returned zero.

flow3r_bsp_imu_read_steps() only copied the count when it caught the watermark interrupt bit set in INT_STATUS_0. Since that register self-clears on read and the watermark only fires every 20 steps, polling almost always read it in a cleared state, so the cached step count was left at its initial 0.

No we read the step counter output register directly  via bmi270_get_feature_data() instead of on the interrupt. That register always holds the live cumulative count.

Also I added imu.step_counter_reset() to zero the counter in the bmi270

Built cleanly via the Docker/ESP-IDF toolchain and flashed to a badge.
Confirmed on hardware: step_counter_read() now increments while walking and step_counter_reset() returns it to 0.
Tested with the example app in the docs (will PR the docs to add the reset funtion)

I left the default watermark granularity on the sensor at 20, I'm not convinced that this is perfect yet in terms of the number matching the actual steps, plus this is a sensor aimed at wrist mounting but at least now if I move about with the badge in my hand the number goes up!

### DIFF
--- a/drivers/tildagon_imu/lsm6ds3/lsm6ds3.c
+++ b/drivers/tildagon_imu/lsm6ds3/lsm6ds3.c
@@ -95,10 +95,20 @@ void lsm6ds3_read_gyro_dps(float *x, float *y, float *z)
  * @brief get step count
  * @param steps pointer for data
  */
-void lsm6ds3_read_steps(uint32_t *steps) 
+void lsm6ds3_read_steps(uint32_t *steps)
 {
     LOCK;
     *steps = _steps;
+    _steps = 0;
+    UNLOCK;
+}
+
+/**
+ * @brief reset step count
+ */
+void lsm6ds3_reset_steps(void)
+{
+    LOCK;
     _steps = 0;
     UNLOCK;
 }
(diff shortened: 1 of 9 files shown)


drivers/tildagon_imu/lsm6ds3/lsm6ds3.c
```diff
@@ -95,14 +95,24 @@ void lsm6ds3_read_gyro_dps(float *x, float *y, float *z)
  * @brief get step count
  * @param steps pointer for data
  */
-void lsm6ds3_read_steps(uint32_t *steps) 
+void lsm6ds3_read_steps(uint32_t *steps)
 {
     LOCK;
     *steps = _steps;
     _steps = 0;
```
ChrisDick, line 102: can you remove this line? that'll make the functionality the same as the bmi. It's only badge team that have the lsm on the prototype baseboards, but it's also our backup if we don't get a sponsor for it.

ChrisDick (dismissed):

drivers/tildagon_imu/lsm6ds3/lsm6ds3.c
sammachin: @ChrisDick ah that make sense I was a little confused why there were 2 sensors, I've removed the reset from the lsm code and just added a guard in the tildagon_imu so it won't try and call the function for that sensor

MatthewWilkes (approved):
Changes have been addressed

PR #343, opened 2026-06-20
Fix sim on Windows

Binney:
# Description

Hi badge team! I tried the sim on Windows and it wouldn't load, you seem to all be Unix users :)

It works with the following 2 changes:
* Mock out `os.statvfs`, which just isn't there on windows
* Add `builtin` as a fallback to the os path resolution mocks

With these 2 it runs fine

{90EAD8DD-B198-472D-B505-4C136779BF1F}

DISCLAIMER
1. I have not tested this on a Unix system because I don't have one. Strongly recommend someone with mac/linux gives it a quick run before merging.
2. AI suggested both these fixes, I am not an expert in this area (not sure what your contribution policy is for this)

### DIFF
--- a/sim/run.py
+++ b/sim/run.py
@@ -68,7 +68,7 @@ sys.path = [
 builtin = BuiltinImporter()
 pathfinder = PathFinder()
 underscore = UnderscoreFinder(builtin, pathfinder)
-sys.meta_path = [pathfinder, underscore]
+sys.meta_path = [pathfinder, underscore, builtin]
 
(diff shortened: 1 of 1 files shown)


MatthewWilkes:
We might need to make statvfs better in future, as it will give an idea of how much free space there is, but this looks fine as is. Good to know it's close to working in windows, I think all windows users (including me) on the badge team use WSL

MatthewWilkes (approved):

PR #347, opened 2026-06-21
Multi-inheritence of buttons and common parents

MatthewWilkes:
This changes the inheritence of buttons to allow them to have multiple inheritence paths, then updates the buttons on the frontboards to point to new common classes to be used across multiple boards. This should make detecting specifically the corner/joystick buttons easier. Also amends the firmware in a couple of places to use this distinction.

### DIFF
--- a/modules/events/joystick.py
+++ b/modules/events/joystick.py
@@ -9,7 +9,8 @@ JOYSTICK_BUTTON_TYPES = {
     "A": Button("A", "Joystick"),
     "B": Button("B", "Joystick"),
     "C": Button("C", "Joystick"),
-    "D": Button("D", "Joystick"),
-    "E": Button("E", "Joystick"),
-    "F": Button("F", "Joystick"),
+    "X": Button("X", "Joystick"),
+    "Y": Button("Y", "Joystick"),
+    "Z": Button("Z", "Joystick"),
+    "SELECT": Button("SELECT", "Joystick"),
 }
--- a/modules/events/joystick.py
+++ b/modules/events/joystick.py
@@ -1,0 +1,15 @@
+from events.input import Button
+
+
+JOYSTICK_BUTTON_TYPES = {
+    "UP": Button("UP", "Joystick"),
+    "DOWN": Button("DOWN", "Joystick"),
+    "LEFT": Button("LEFT", "Joystick"),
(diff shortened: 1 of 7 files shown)


MatthewWilkes:
@ChrisDick This is inspired by some of the earlier code you had in your PR, does this make some of the problems easier?

modules/firmware_apps/app_store.py
```diff
@@ -490,17 +491,17 @@ def __init__(self, install_handler: Callable[[str], Any], app: app.App):
     def _handle_buttondown(self, event: ButtonDownEvent):
         kbd_button = event.button.find_parent_in_group("Keyboard")
 
-        if BUTTON_TYPES["UP"] in event.button:
```
ChrisDick, line 532: does this not mean that the joystick can't be used in the app store, but can be used on the other firmware apps? the dialog change makes sense.

modules/events/joystick.py
```diff
@@ -0,0 +1,15 @@
+from events.input import Button
+
+
+JOYSTICK_BUTTON_TYPES = {
+    "UP": Button("UP", "Joystick"),
+    "DOWN": Button("DOWN", "Joystick"),
+    "LEFT": Button("LEFT", "Joystick"),
+    "RIGHT": Button("RIGHT", "Joystick"),
+    "A": Button("A", "Joystick"),
```
ChrisDick, line 9: Not sure on the button names, I can see why you went for 6 buttons, but should we go a,b,c,x,y,z and select?  this would cover a good few retro joypads or do you want to control the menu from the joypad?

ChrisDick (changes requested):

ChrisDick:
yes, this should allow the same idiom to be used for both generic and specific buttons.

modules/firmware_apps/app_store.py
ChrisDick: ignore this, the Codeinstall had been chopped off the diff.

MatthewWilkes: I see you resolved this, so I think you spotted what this handler is for, but to clarify for anyone else: This is the code input handler, which we're specifically using locations around the hexagon for, not directions.

ChrisDick (approved):

PR #350, opened 2026-06-23
Test the simulator in builds

Jonty:
Because we keep breaking either the main code, or the simulator.

This isn't testing the web emulator yet.


MatthewWilkes (approved):
Nice. Would be good to comment the screenshot in future, maybe?

PR #351, opened 2026-06-23
Placeholder fake for 2026 c support code

MatthewWilkes:
Adds a nonfunctional placeholder for the 2026 frontboard's C driver. If we want to add this to the sim properly, we should probably get on that, but this fixes boot.

Also regenerates the ctx wasm to embed the latest version of the camp font.


MatthewWilkes:
FYI @Jonty and @pikesley

Jonty (approved):

PR #352, opened 2026-06-23
Add support for i2s PDM to mpy

joenash:
# Description

This patch adds support for I2S PDM to micropython, as implemented by this (unmerged) 2024 PR to micropython: https://github.com/micropython/micropython/pull/14176

# Motivation
The ESP32-S3 supports PDM via its I2S peripherals, and ESP-IDF has libs for it, but micropython does not. The aforementioned PR added support for it, but appears to have gotten stuck in API bikeshedding hell. Given PDM is supported through I2S on this board, I think it being part of `machine.I2S` was quite an appropriate solution.

One use case of this is PDM microphones like the SPH0641LU4H-1. I'm using this in a hexpansion for bat detection. I have confirmed this patch is working, and allows me to use the microphone.

I am entirely new to micropython, ESP32, and firmware, apologies for any horrors!

### DIFF
--- a/patches/micropython-i2s-pdm.diff
+++ b/patches/micropython-i2s-pdm.diff
@@ -1,0 +1,266 @@
+diff --git a/docs/library/machine.I2S.rst b/docs/library/machine.I2S.rst
+index 84edb94e7..8832885c4 100644
+--- a/docs/library/machine.I2S.rst
++++ b/docs/library/machine.I2S.rst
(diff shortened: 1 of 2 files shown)


MatthewWilkes (approved):
We were trying to keep the patches for the micropython changes together, but we're accumulating enough independent ones now that I'm rethinking that. Merging this, but let's talk about it as the badge team at some point that isn't squeaky bum time.

PR #354, opened 2026-06-24
Ctx patches v2

b0xcat:
# Description

New ctx patches from https://ctx.graphics/tildagon/, plus a workaround for the horizontal lines through text sometimes that hopefully doesn't break anything else


MatthewWilkes (approved):
pippin knows best

PR #355, opened 2026-06-25
Sim: Add a fake for micropython.const()

mbooth101:
# Description

This is used to optimise constants in micropython, but is not present in cpython. Adding this fake allows apps that use this optimisation to run in the pygame simulator.


MatthewWilkes (approved):
Thanks

Issue #353, opened 2026-06-24
Crashing Yes/No dialog


pikesley:
When attempting to quit an app using the `YesNoDialog`, the app crashes hard with:

```
Registered async event handler for ButtonDownEvent: EMFCave - _handle_buttondown
Removed event handler for ButtonDownEvent: EMFCave - _handle_buttondown
[Screen reader] 0
Traceback (most recent call last):
  File "main.py", line 62, in <module>
  File "system/scheduler/__init__.py", line 261, in run_forever
  File "asyncio/core.py", line 1, in run_until_complete
  File "asyncio/core.py", line 1, in run_until_complete
  File "asyncio/core.py", line 1, in run_until_complete
  File "system/scheduler/__init__.py", line 257, in _main
  File "asyncio/funcs.py", line 1, in gather
  File "asyncio/core.py", line 1, in run_until_complete
  File "system/eventbus.py", line 101, in run
  File "asyncio/funcs.py", line 1, in gather
  File "asyncio/core.py", line 1, in run_until_complete
  File "app_components/dialog.py", line 81, in _handle_buttondown
TypeError: 'NoneType' object isn't iterable
MicroPython v1.28.0-dirty on 2026-06-24; Tildagon with ESP32S3
```

and the Tildagon locks and has to be rebooped

Version: `main`, `4c954f1`

**To Reproduce**

With a running shell on the badge:

1. Run an app that uses the `YesNoDialog`, e.g. https://apps.badge.emfcamp.org/apps/23340441/ 
2. Do whatever you need to get it to pop-up (in this case, just crash the ship)
3. Attempt to select `No`
4. See it crash

**Expected behavior**
It should just quit, and this has been working fine up until now

TechCabin:
found the issue inside modules/app_components/dialog.py

    async def _handle_buttondown(self, event: ButtonDownEvent):
        if BUTTON_TYPES["CANCEL"] in event.button:
            self._cleanup()
            self._result = False
            if self.no_handler is not None:
                result = **await** self.no_handler()
                if hasattr(result, "pend_throw"):
                    await result

        if BUTTON_TYPES["CONFIRM"] in event.button:
            self._cleanup()
            self._result = True
            if self.yes_handler is not None:
                result = self.yes_handler()
                if hasattr(result, "pend_throw"):
                    await result

CANCEL contains await, but self.no_handler defaults to None

def __init__(self, message, app, on_yes=None, **on_no=None**):
        self.open = True
[...]

PR #356, opened 2026-06-25
Bugfix/fix crashing dialog

pikesley:
# Description

This addresses the Crashing Dialogs bug identified in #353

### DIFF
--- a/modules/app_components/dialog.py
+++ b/modules/app_components/dialog.py
@@ -77,8 +77,8 @@ class YesNoDialog:
         if BUTTON_TYPES["CANCEL"] in event.button:
             self._cleanup()
             self._result = False
-            if self.no_handler() is not None:
-                result = await self.no_handler()
+            if self.no_handler is not None:
+                result = self.no_handler()
                 if hasattr(result, "pend_throw"):
                     await result
 
--- a/modules/app_components/dialog.py
+++ b/modules/app_components/dialog.py
@@ -77,7 +77,7 @@ class YesNoDialog:
         if BUTTON_TYPES["CANCEL"] in event.button:
             self._cleanup()
             self._result = False
-            if self.no_handler is not None:
+            if self.no_handler() is not None:
                 result = await self.no_handler()
                 if hasattr(result, "pend_throw"):
                     await result


TechCabin:
This fix may be affecting the shutdown procedure

modules/app_components/dialog.py
```diff
@@ -77,7 +77,7 @@ async def _handle_buttondown(self, event: ButtonDownEvent):
         if BUTTON_TYPES["CANCEL"] in event.button:
             self._cleanup()
             self._result = False
-            if self.no_handler is not None:
+            if self.no_handler() is not None:
                 result = await self.no_handler()
```
MatthewWilkes: Don't we want to remove this await instead, as it is awaiting twice?

pikesley:
> This fix may be affecting the shutdown procedure

Oh, that may explain some Stuff I've been seeing on my badge today

modules/app_components/dialog.py
pikesley: Entirely plausible, I have only ever pretended to understand async and await and all that

pikesley: Aha, I removed this and it seems to work

TechCabin:
so we cannot call self.no_handler, because it defaults to None

This causes an issue during the shutdown procedure

Creating app firmware_apps.poweroff.PowerOff...
Registered async event handler for ButtonDownEvent: PowerOff - _handle_buttondown
Removed event handler for ButtonDownEvent: PowerOff - _handle_buttondown
inside dialog.py _handle_buttondown
  File "main.py", line 55, in <module>
  File "system/scheduler/__init__.py", line 261, in run_forever
  File "asyncio/core.py", line 1, in run_until_complete
  File "asyncio/core.py", line 1, in run_until_complete
  File "asyncio/core.py", line 1, in run_until_complete
  File "system/scheduler/__init__.py", line 257, in _main
  File "asyncio/funcs.py", line 1, in gather
  File "asyncio/core.py", line 1, in run_until_complete
  File "system/eventbus.py", line 101, in run
  File "asyncio/funcs.py", line 1, in gather
  File "asyncio/core.py", line 1, in run_until_complete
  File "app_components/dialog.py", line 81, in _handle_buttondown
TypeError: 'NoneType' object isn't callable
MicroPython v1.28.0-dirty on 2026-06-26; Tildagon with ESP32S3
Type "help()" for more information.
>>>

TechCabin:
Confirmed this works and does not break the shutdown:

    async def _handle_buttondown(self, event: ButtonDownEvent):
            self._cleanup()
            self._result = False
            if self.no_handler is not None:
                result = self.no_handler() #removed await
                if hasattr(result, "pend_throw"):
                    await result

MatthewWilkes (approved):

PR #361, opened 2026-06-27
Catch bad eventbus handlers and end the app

benclifford:
Before this PR, the firmware crashes out to micropython repl when an eventbus handler raises an exception.

~~(draft because I haven't thought about this very hard, but its been in my patch stack for 5 months without causing me trouble)~~

### DIFF
--- a/modules/system/eventbus.py
+++ b/modules/system/eventbus.py
@@ -2,6 +2,10 @@ import asyncio
 
 from async_queue import Queue as AsyncQueue
 from perf_timer import PerfTimer
+from system.scheduler.events import RequestStopAppEvent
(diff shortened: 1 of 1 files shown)


benclifford:
I just looked at this again less tired, I think this is ok to be ready for review. I was originally wondering about if the behaviour on failure was right - but it is the same as everywhere else.

b0xcat:
LGTM! Definitely a good addition to have if we allow arbitrary dicts and custom events to be sent through the eventbus like here https://github.com/emfcamp/badge-2024-software/pull/372

Would be great to have some tests to verify the correct behavior too, although we don't really have a test setup until that is merged.

ChrisDick (approved):

PR #364, opened 2026-06-27
Update default SSID for 2026 (and, hopefully, onwards)

lukegb:
# Description

The current default SSID is old; although people who have the badge by now have probably set their own SSID, we should probably make sure that people who get badges for the first time actually have it connect to the wifi OOTB.


MatthewWilkes (approved):
Thank you, this was on the todo list, but nobody had gotten to it yet.

PR #365, opened 2026-06-27
Support loading hexpansion drivers from flash

MatthewWilkes:
Some hexpansions have small EEPROMs, this change allows updates to be stored to the flash if there is no filesystem present when updating a hexpansion. Those files will then be loaded when a matching hexpansion is inserted, assuming no FS is present.


modules/system/scheduler/utils.py
```diff
@@ -0,0 +1,21 @@
+from . import scheduler
+
```
ChrisDick: is this file part of another pr?

MatthewWilkes: I was just being lazy and including it, but I should split it.

hughrawlinson (approved):

PR #366, opened 2026-06-28
add randomness to badge detection

ChrisDick:
overwrite tildagon message with a random number and use this to check the badge isn't talking to itself. move the send id helpers to c to let them use the random.

# Description

this stops the badge thinking it's connected to another badge when in lanyard mode.


drivers/tildagon_power/tildagon_power.c
```diff
@@ -474,8 +477,24 @@ void device_pd ( event_t event )
                 && ( usb_in.pd.vendor.vendor_data[7] == tildagon_message[7] )
             )
             {
-                push_event( MP_POWER_EVENT_BADGE_AS_DEVICE_ATTACH );
-                badge_as_device = true;
+                if (
+                    ( usb_in.pd.vendor.vendor_data[12] == tildagon_message[12] )
+                    && ( usb_in.pd.vendor.vendor_data[13] == tildagon_message[13] )
+                    && ( usb_in.pd.vendor.vendor_data[14] == tildagon_message[14] )
+                    && ( usb_in.pd.vendor.vendor_data[15] == tildagon_message[15] )
```
MatthewWilkes: Given it's `esp_fill_random( &tildagon_message[12], 8 );`, shouldn't this be 8 bytes being checked? Would `memcmp(usb_in.pd.vendor.vendor_data + 12, tildagon_message + 12, 8)` be easier? (I've not tried that)

drivers/tildagon_power/mp_pd.c
```diff
@@ -213,13 +213,31 @@ static mp_obj_t host_send_dbl_prime_msg( mp_obj_t self_in, mp_obj_t mp_data )
 
 static MP_DEFINE_CONST_FUN_OBJ_2( host_send_dbl_prime_msg_obj, host_send_dbl_prime_msg);
 
+static mp_obj_t device_send_badge_id( mp_obj_t self_in )
+{
+    fusbpd_vendor_specific( &usb_in.pd, tildagon_message, 5 );
```
MatthewWilkes: 5 is because it's 20 bytes, at 4 bytes per object, right? Would `sizeof(tildagon_message) / 4` make it look less magical?

ChrisDick:
I think that this is only connecting to other badges at higher battery voltages, doing some testing, but it's going to take a while to discharge. the reduction in current limit is to help with this

drivers/tildagon_power/tildagon_power.c
ChrisDick: I've done this to all of the comparisons. I'm trying to only use the minimum amount of the message to leave space for future, have reduced it to a random fill of 4 bytes.

ChrisDick:
this still wip. I get a notification on first connect of the lanyard, then it's ok. This does improve it, but it's not there yet. 

I've tested the connection down to 0.6% battery on host. I think the battery droop is still causing an issue when charging another badge. That can be another pr if we want it as it will stop being able to charge other badges and we might want to do that.

ChrisDick:
the disconnect event wasn't being generated because the PMIC check had to wait for a usb in interrupt, moved this to the main task as it fits nicely with the 1s timeout on the queue. This still occasionally doesn't disconnect the device, but will allow another reconnection.

MatthewWilkes (approved):

PR #367, opened 2026-06-28
Add minimal autoexec.bat support

benclifford:
This is implemented in the launcher rather than calling directly in the scheduler so that the launcher is aware of launched apps, for re-entry off the launcher menu later.

This code matches against the name of the app, because that seems more user friendly than the class name.

This code matches a substring of the app name, so that you can launch "Duck Facts" when the app is actually called "Duck Facts [[EMF UNICODE VIOLATION DUCK :duck: ]]". The first matching app is launched, in whatever order the launcher built its menu.

A couple of possible use cases from IRC:

* less patching of core for the wasm emulator port to autoinstall apps: with this patch you can make an emulator-startup app, and have that run on startup

* running spyware on your kids badges automatically at startup

In IRC we talked about running multiple apps and running arbitrary Python. This does not directly support that.

For arbitrary Python, implement your own app (that's what apps are for!) and start that app using this mechanism.

For starting multiple apps, there isn't an exposed interface to launch apps (e.g. over eventbus) that can then be re-entered from the launcher. Fixing up that was more than I wanted to bite off in this PR. See PR #321 for related work for hexpansion apps.

### DIFF
--- a/modules/system/launcher/app.py
+++ b/modules/system/launcher/app.py
@@ -15,6 +15,7 @@ from system.scheduler.events import (
 )
 from system.notification.events import ShowNotificationEvent
 from app_components.background import Background as bg
+from app_components.tokens import symbols
 
 APP_DIR = "/apps"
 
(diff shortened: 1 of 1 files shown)


Jonty:
It might be nice if this showed a quick splash saying that autoexec.bat is running and what the app name is (and uses the new bat character!), and shows an error screen for a second if the app isn't installed, just to make it obvious what's happening and if it isn't working.

You can nick the code for the autostart splash screen/error from: https://github.com/emfcamp/tildagon_web_emulator/blob/main/fakes/autostart.py

benclifford:
oh right now its not even going to log a console error if theres no app found. wups.

benclifford:
I don't really want to be messing with the launcher UI right now to add new screens. (if I was putting Big Effort in here, I'd rather move to an out-of-launcher event driven system that probably interacts in some way with PR #321 hexpansion apps too).

If the use cases are end-users putting their own autoexec.bat on (rather than some evil app doing so) then it should be obvious to many people why the app is starting up. And apps usually start fast -- this isn't directly the web emulator "download a load of stuff" case, where a "please wait" screen would help. So I'm not too saddened by not doing a splash screen.

For similar UI avoiding reasons, I just made the error messages appear using the notification message and added in an app-not-found case. Which displays the bat!

image


I'm open to working on a more elaborate systemd-tildagon for EMF 2028.

ChrisDick (approved):

PR #369, opened 2026-06-29
Add esp-now service and eventbus integration

b0xcat:
# Description
This PR adds a shared esp-now service so apps don't each have to grab the radio and set it up properly.

It:
- is hooked up to the eventbus for easy message listening by multiple apps
- allows filtering incoming messages using the `predicate` argument
- allows both sync and async message sending
- automatically manages wifi settings and power management, wifi power saving can cause dropped messages. We disable power saving when anyone is listening for esp-now messages and re-enable it once no one listens anymore.

Also included:
- Buttons.pressed(), a convenience latched read that fires once per press instead of every frame.
- An example "ESPNow ping" app (CONFIRM broadcasts a hello, screen shows last received + RSSI/channel/timestamp). It is in the menu for easy testing, happy to remove it from there before merging.


modules/firmware_apps/espnow_ping.py
```diff
@@ -0,0 +1,58 @@
+import app
```
ChrisDick, line 1: I think this would be better in modules/firmware apps with the other examples.

ChrisDick (changes requested):

ChrisDick:
I agree to taking the app out of the launcher, otherwise looks good to me

hughrawlinson:
@b0xcat this sounds like a capability to me!

modules/firmware_apps/espnow_ping.py
b0xcat: Sure, I moved it

b0xcat:
Then this is ready to go hopefully!

ChrisDick (approved):

PR #370, opened 2026-06-29
revert pd setup

ChrisDick:
# Description

change to test, using CI to build and let people get the bin from the action. not for merge yet.


MatthewWilkes:
From short initial testing, this doesn't appear to have an impact on USB connectivity

ChrisDick:
Are you still seeing the mpremote problem?

MatthewWilkes:
Yep. It isn't specifically mpremote though, it just doesn't get recognised by the OS. The PD does trigger, it is the data connection that intermittently fails. Without any feedback in Windows, hence why I keep meaning to capture in Linux.

ChrisDick:
ok, I'll close this pr and raise an issue to keep track of this.

PR #372, opened 2026-07-01
Add CustomEvent and make eventbus more flexible

b0xcat:
# Description

After the discussion about ad-hoc events and capabilities today, it seemed good to allow subscribing to events with only a string instead of a class definition. 

People were already using dicts and subscribing to dict events, which puts the burden of filtering for the events they are interested in on their app. This should be the responsibility of the eventbus, so this PR modifies it to make it more flexible and allow subscribing based on strings:

```python
eventbus.on("somethinghappened", dothething, app)
```

will call `dothething` for `CustomEvent`s with `type = somethinghappened`:

```python
eventbus.emit(CustomEvent(type="somethinghappened", data={"whatever": "youwanttoputhere"})
```

or `dict`s with an entry `yourdict["type"] = "somethinghappened"`:

```python
eventbus.emit({"type": "somethinghappened", "otherthing": "hi")
```

You can also still subscribe to classes (even `dict` to just get all dict events) like before

This seems like a good way to sidestep the issues with needing to share event classes / PRing them into the firmware etc.

Let me know what you think!

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,19 @@
+name: Unit tests
+on:
+  pull_request:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+jobs:
+  unittest:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+      - name: Setup python
+        uses: actions/setup-python@v5
+        with:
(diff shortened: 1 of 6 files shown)


MatthewWilkes (approved):
This looks good, but needs a rebase

MatthewWilkes (approved):

Issue #371, opened 2026-06-29
USB connectivity


ChrisDick:
**Describe the bug**
usb doesn't reliably connect to laptop. seen on windows and mac.

Version: [for example, v2.0alpha]

**To Reproduce**
 plug in badge to a usb pd port with a C-C cable.

**Expected behavior**
appears as a usb device

**Additional context**
seems to work with a usb A-C cable.
#370 didn't fix it.

ChrisDick:
Further investigation has shown that this issue was introduced in v1.9.1 with the enabling of PD comms. Specifically with the enabling of auto CRC response on the fusb302b. The good CRC response is sent by the fusb302b when a received message matches the CRC of the message. The response is generated and an interrupt generated for us to read and respond. 

This allows us to receive the pd supply capabilities message and request up to 3A instead of the 1.5A limit without PD. This functionality is released and working. There is a change in v2.a that also uses this to receive messages from another badge without continuous polling. As this issue started with v1.9.1 we can eliminate the change in v2.a. We can also eliminate any changes with esp idf and micropython brought in with v2.0a. 

The issue is this: when a dual role port such as a thunderbolt or a display port connects to the badge it  successfully negotiates the additional current with the badge, then interrogates the badge with a discover identity request. The fusb302b responds with it's auto generated good CRC response. The badge doesn't respond because I wasn't expecting the USB inquisition.

I have tried NACKing the request and sending a dummy response, both had no effect. Reading up on this I found that the contents of the discover identity include a header of what the badge is capable of, a certificate issued by the USB-IF, a product id and revision. The USB host uses these additional fields to determine if it's a safe combination and I'm assuming the badge needs to respond with a safe pairing before the OS will allow the USB data to be connected.  

No idea what a safe pairing would be.

PR #376, opened 2026-07-04
Usb connection fix

ChrisDick:
# Description

fix for #371 

on reception of the discover id request message, disconnect the pull downs, wait a bit and don't enable pd on the reconnect. This causes the thunderbolt port to allow usb data to connect, but only allows 1.5A charge current from the port.


ChrisDick:
testing on danny's machine, this doesn't trigger the reset, so there is something else missing

MatthewWilkes:
@ChrisDick I am interpreting this as meaning this fix isn't ready for testing, is that right? Do we want to disable that increased PD behaviour for now?

ChrisDick:
a bit more testing, probably on camp when i can grab more traces. I don't think we should disable the PD, it looks like this only effects thunderbolt 4 ports, and an c-a adapyer and an a-c  cable work on those ports.

MatthewWilkes:
I am just being selfish in that I have ~~none~~ only one semi functional one of those things, and all my laptop ports trigger this

cachedglyph:
Having hit this issue and ended up spending quite a while trying to figure out what was up, if it's not being merged before the event, it may be worth including some warnings in docs / on the flasher tool perhaps until it's fixed.
If it's anything with a thunderbolt 4 port specifically, then that's quite a lot of Mac laptops for example out there (Macbook Pros since 2021 and Macbook Airs since 2025).

Especially for folks who previously it worked with their laptop in 2024, then updated the firmware on the badge to find it now doesn't work, with odd behaviour. I assumed initially it was a cable issue before trying on another machine.

walkerdanny:
For what it's worth, USB C works for me on one laptop but not another, newer laptop.

Both are Thinkpads, one is an X1 Carbon Gen 9 with Thunderbolt 4 (working) and the other is a P1 Gen 7 with Thunderbolt 4 (not working). The X1 Carbon works completely as expected, but the P1 doesn't show up as a COM port.

Both running Windows 11 Pro, tried with multiple cables etc.

MatthewWilkes:
@ChrisDick I just tried it, and it seems to fix it for me.

MatthewWilkes (approved):
It improves things, even if it doesn't fix every case

ChrisDick:
Going to merge this in as it is helping, but may not fix for everyone. I've had a few thoughts about what else we could do, but we can discuss tomorrow.

PR #378, opened 2026-07-05
Add configurable Menu Highlight Colours

Coedy:
**This is a new version of #330 following the 2026 theme work that recently got merged in.**
=============================================================================
If you install multiple long named apps from the store its hard to tell which one you are highlighting as there isn't much room for the app name to enlarge.
Focused menu items render in a user-selectable highlight colour, making it much easier to see which item is selected, especially for long app names.

image of badge simulator homescreen with custom menu colour

Works on main menu, sponsors, & any level within the appstore as uses the deault methods for navigating through lists.

Six colours available: Yellow, Pink, White, Green, Orange, Blue. Theme sets is to the default of the badgefront (white for 2024 the Sim uses)
Setting persists reboots/reloads of the Sim via the existing settings system and can be changed in the Settings app under 'Menu highlight' section.

Also fixes a simulator bug (hence the sim/run.py change) where bare root paths like /settings.json were not redirected to /tmp/sim/, preventing settings from persisting across simulator sessions and crashing the sim when trying to exit after a change (Permissions Error on logs).

Only tested on the Sim. 

Created using Claude as I had tried this twice previously and wasn't able to get it working. Also had to use Claude to work out all the changes in the 2026 theme stuff.


Coedy:
Re-committed following notifications of ruff errors.

Issue #326, opened 2026-06-02
Add App description and author to app store


Jonty:
Currently when browsing the app store selecting an app immediately installs it, we should instead show a screen with the app description, author, and version, and an option to install it.

We may need to make the description auto-scroll or something?

andypiper:
Currently you can press B to get the description (and DOWN to scroll that, but it never re-centres for the next one), but yeah I think adding all of this would be helpful. Maybe small icons by each button to indicate info and install.

PR #379, opened 2026-07-06
Add App details page to AppStore

Coedy:
## Overview:
I kept finding myself accidentally installing items instead of pressing B to read the descriptions.
I also noticed it was raised as Issue  #326 
So despite my lack of technical knowledge I decided to try it (with help 🤖).

Added app details page and updated menu interactions. (eg: B button no longer displays the basic description)

Users can now browse apps via:
AppStore > Browse > Category > App name > Details page > Install button

## Video:
[Screencast_20260706_140454.webm](https://github.com/user-attachments/assets/b0b225a4-ae88-4f19-b34a-6eee6058d3a6)

## Technical changes:

All changes are contained in `modules/firmware_apps/app_store.py`.

- Adds a new `app_details` state to `AppStoreApp`'s state machine, sitting between `available_menu` and `installing_app`. Selecting an app in the browse menu no longer sets `to_install_app` directly; it stores the selected index entry in `details_app` and transitions to `app_details`. `to_install_app` is only set (and the existing `installing_app` state entered) when the user confirms from the detail page, so the download/install path itself is unchanged.

- "Install now" renders in `label` white until max scroll is reached, then in the `active_menu_item` theme token to indicate it is selected. (so this also works with PR #378 and the new Themes in the 2026 badgefront hence the yellow text colour in the video)
Button mapping in this state: UP/DOWN scroll, CONFIRM installs only when the install row is selected (otherwise a no-op), CANCEL returns to the list. The browse menu's list position is preserved across the round trip (reset on category change).

- Removes `info_items` from the browse menu's `Menu` constructor, which disables the Menu widget's RIGHT-button (B) description overlay in the app store; the detail page supersedes it. The Menu widget itself is unmodified and other `info_items` users across the badge are unaffected.

**NOTE for transparency:**
**Created using Claude,** 
Have tried it in the simulator successfully but not directly on the badge.


Coedy:
Re-committed following notifications of ruff errors.

PR #382, opened 2026-07-07
Fix pin assignment for LS 5 on hex port E

DanNixon:
# Description

This was seemingly wrong, updated to match the schematic.


ChrisDick (approved):
good spot.

PR #383, opened 2026-07-07
Update menu items in AppStoreApp to conditionally show uninstall option

ntflix:
In the App store, if there are no apps installed, selecting "Uninstall" crashes the app as it attempts to show a zero-length Menu. This commit only shows the "Uninstall" option in the App store if `len(list_all_apps()) > 0`.


ChrisDick (approved):
Thanks

PR #391, opened 2026-07-11
feat: use containerized app store

hughrawlinson:
Also, a small fix to allow for multi-category apps

### DIFF
--- a/modules/app_components/menu.py
+++ b/modules/app_components/menu.py
@@ -120,6 +120,11 @@ class Menu:
         )
 
     def _calculate_max_focussed_font_size(self, item, ctx):
+        if not isinstance(item, str):
+            print(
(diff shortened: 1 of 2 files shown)


MatthewWilkes (approved):
Works for me

hughrawlinson:
I'm *so* confused now - this works fine because it grabs the listing from the new url, which contains the new tarball links (we log the download and then 302 to another endpoint pull the tarball through a cache so we don't rely on github being up). But, when I modify main to just rewrite the tarballUrl to the new format, it fails to follow the redirect because the location header is lowercase intstead of captitalized. But why would that work on this branch??

hughrawlinson:
anyway - I've added an upper case Location header in the app store, so that's that solved. this can be merged and the app store can set the tarballurls in the `/demo_apis/apps.json` listing api to the new endpoint and we should then get telemetry from all installs.

PR #392, opened 2026-07-11
Make PatternEnable/Disable events apply to backleds

### LINKED ISSUE #387: conflict between emote events and user apps for back leds
**Describe the bug**
the emote events pr brought in a back led manager, this is always controlling the back leds and isn't disabled with eventbus.emit(PatternDisable()) causing leds to flicker when a user app uses the leds through _tildagonos

Version: e96b850ed4547ad0ce42fa92684f29b304ce6390 of main

**To Reproduce**
an app lighting all 18 LEDs, but the rear ones are flickering badly

**Expected behavior**
no flicker.

**Additional context**
the emote events pr brought in a back led manager, this is always controlling the back leds and isn't disabled with eventbus.emit(PatternDisable())

benclifford:
This means that apps taking control of the LEDs now also have the opportunity to do fun stuff with the back-of-board LEDs.

Fixes #387

I manually tested:

 * writing to back LEDs flickers as described in #387, as the app and backled managers are both writing. This is the same flicker you get when writing to the front LEDs when patterns are enabled.

 * emote events show interleaved in the flicker

 * after a PatternDisable: there is no flicker. emote events are not rendered on the backleds at all.

 * after a PatternEnable: the behaviour goes back to the flickering behaviour

I also had a feeling that this flickering behaviour would have manifested before the backled manager was introduced in PR #244 if the "mirror patterns" setting was enabled, and I checked that this was indeed the case.

### DIFF
--- a/modules/system/backleds/app.py
+++ b/modules/system/backleds/app.py
@@ -5,6 +5,7 @@ from app import App
 from events.emote import EmotePositiveEvent, EmoteNegativeEvent
 from system.eventbus import eventbus
 from system.hexpansion.events import HexpansionInsertionEvent, HexpansionRemovalEvent
+from system.patterndisplay.events import PatternDisable, PatternEnable
 from tildagonos import tildagonos, led_colours
 
 # note that event.port is indexed from 1-6,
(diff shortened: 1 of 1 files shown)


ChrisDick (approved):
thanks

PR #393, opened 2026-07-11
feat: use new tarball download endpoint 

hughrawlinson:


hughrawlinson:
we won't merge this, it's just for testing that the new tarball endpoint works ok with the http stack before I switch it for everyone. It looks like it does _not_ work, which is a bit of a bummer.

Issue #389, opened 2026-07-10
Display does not persist


pikesley:
Several apps which rely on the display persisting are broken. [Hilbert](https://apps.badge.emfcamp.org/apps/20242041/) and [Barnsley Fern](https://apps.badge.emfcamp.org/apps/24031134/) are good examples

Version: `03ee2b1`

**To Reproduce**

1. Install Barnsley Ferm
2. Start the app
3. Watch it throw some dots about

**Expected behavior**

It's supposed to build up an image by drawing distinct dots onto the canvas at each iteration. However it seems like it's clearing the canvas every time so we just get a load of dots crawling about. Also the background is supposed to be a defined colour but it looks like it's black every time

**Additional context**

This works fine on `2.0.0alpha5`

Issue #390, opened 2026-07-10
PNG display is unreliable


pikesley:
Apps that display PNGs are very unreliable. This is all a bit vague and hand-wavy, but for example [Valued Goose](https://apps.badge.emfcamp.org/apps/33240411/) runs fine the first time, but if you come out and then go back in (or into another app that uses PNGs) the badge behaves like it's running but no image is displayed, the menu remains on the screen. And sometimes the badge crashes hard

Version: `03ee2b1`

**To Reproduce**

1. Install Valued Goose
2. Start it
3. Exit (button F)
4. Go back in
5. Eventually the image will fail to show

**Expected behavior**
We should see an image of a Valued Goose

**Additional context**
This works fine on `2.0.0alpha5`

andypiper:
I have the same issue with Brick Space FWIW. I can get through a few cycles of image changes, but then it - somewhat inconsistently - shows nothing.

With Duck Facts, the Duck Hunt feature (grabbing duck images from a random duck API for display, the C button) does not show images with the HEAD builds of the firmware.

MatthewWilkes:
After a few attempts at bisecting, I think this is caused by 1550f46bd824c72e83882b86e4875c8686a6c2dc, I am going to do a PR to disable drawlist mode

pikesley:
I think this came back: some of my PNG-based apps appear (in the console) to be running, but nothing happens on the screen

PR #397, opened 2026-07-11
Disable drawlist mode

MatthewWilkes:
We are seeing some bugs related to image rendering, such as #390. From bisecting, I believe that drawlist mode may be the reason. This PR disables it, while leaving the code to implement it in place.


MatthewWilkes:
This also appears to fix #389

pikesley:
Agreed, this seems to fix #389 and #390

ChrisDick (approved):

PR #405, opened 2026-07-13
Compact representation for capabilities

MatthewWilkes:
This adds a compact representation for the capabilities function, for hexpansion eeproms

### DIFF
--- a/modules/system/capabilities/utils.py
+++ b/modules/system/capabilities/utils.py
@@ -203,6 +203,20 @@ def list_all_requirements():
     return result
 
 
+def get_manifest_from_compact_app_format(app_obj):
+    identifiers = getattr(app_obj, "CAP", [])
+    identifiers = [
+        identifier.replace(
+            "@", "https://tildagon.badge.emfcamp.org/capabilities/registry/"
+        )
+        for identifier in identifiers
+    ]
+    capabilities = [
+        {"capability": {"identifier": identifier}} for identifier in identifiers
+    ]
+    return {"metadata": {"providedCapabilities": capabilities}}
+
+
 def app_sort_key(app):
     try:
         return getattr(app, "priority", 1)
--- a/modules/firmware_apps/hexpansionfw.py
+++ b/modules/firmware_apps/hexpansionfw.py
@@ -30,8 +30,8 @@ from system.notification.events import ShowNotificationEvent
 DEFAULT_REPO = (
     "https://github.com/emfcamp/hexpansion-firmwares/releases/download/latest/"
 )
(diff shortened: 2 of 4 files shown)


MatthewWilkes:
Also fixes a capitalisation bug in hexpansion firmware ota

modules/system/capabilities/utils.py
```diff
@@ -203,6 +203,20 @@ def list_all_requirements():
     return result
 
 
+def get_manifest_from_compact_app_format(app_obj):
+    identifiers = getattr(app_obj, "CAP", [])
```
hughrawlinson, line 207: The disadvantage of this option is that to know what capabilities are supported provided by an hexpansion app, we have to evaluate the python code. Maybe it's fine, I don't know a better solution.

modules/system/capabilities/utils.py
```diff
@@ -203,6 +203,20 @@ def list_all_requirements():
...
+    identifiers = [
+        identifier.replace(
+            "@", "https://tildagon.badge.emfcamp.org/capabilities/registry/"
```
hughrawlinson, line 210: I guess this means that people who choose to publish their capabilities elsewhere can't have an @ symbol in their url, that's fine right? Is @ a valid URL character anyway?

modules/system/capabilities/utils.py
```diff
@@ -203,6 +203,20 @@ def list_all_requirements():
...
+        )
+        for identifier in identifiers
+    ]
+    capabilities = [
+        {"capability": {"identifier": identifier}} for identifier in identifiers
+    ]
+    return {"metadata": {"providedCapabilities": capabilities}}
```
hughrawlinson, line 217: Will we allow these apps to support/require capabilities?

modules/system/capabilities/utils.py
MatthewWilkes: There is never a case where we have a hexpansion app without doing that, though, as the app is instantiated as part of the insertion event

modules/system/capabilities/utils.py
MatthewWilkes: Well, these aren't necessarily URLs, right? We could make it just replace the first character, but it felt like a fair constraint at this stage to get it working.

modules/system/capabilities/utils.py
MatthewWilkes: I think that we might want to add that later, when/if we have automatic launching of apps that offer required capabilities, but that's not happening this year, and I'm inclined to think that people will do it from code rather than declaratively.

ChrisDick (approved):

PR #406, opened 2026-07-14
Provide a "terminate this app" API

pikesley:
# Description

Actually remove the app from the scheduler rather than `minimise` it. I've been using this in some of my apps and it feels like it might be more generally useful.
 
Also, and this might just be me, but it's easy to forget to restore the LED pattern and return them in whatever state your app left them.

### DIFF
--- a/modules/app.py
+++ b/modules/app.py
@@ -4,6 +4,9 @@ import time
 from system.eventbus import eventbus
 from system.scheduler.events import RequestForegroundPopEvent
 
+from system.scheduler.events import RequestStopAppEvent
+from system.patterndisplay.events import PatternEnable
+
 
 class App:
     def __init__(self):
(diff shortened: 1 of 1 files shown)


MatthewWilkes:
Could you put it on the app class, at https://github.com/emfcamp/badge-2024-software/blob/main/modules%2Fapp.py#L52 alongside minimise?

pikesley:
> Could you put it on the app class, at https://github.com/emfcamp/badge-2024-software/blob/main/modules%2Fapp.py#L52 alongside minimise?

Ah yes of course a much more sensible location :+1:

benclifford:
in some comment that isn't appearing for me any more, I mentioned to @pikesley that this restores patterns even if some other app is the one that disabled patterns. in different behaviour, an app that crashes and is terminated by other handling gets left with no pattern enabled, I think. I think these are both wrong behaviours, and I wanted to note that here as a thing to expect, without expecting this PR #406 to change -- maybe its something for thinking about longer term elsewhere for apps releasing resources automatically.

MatthewWilkes:
I think that's a fair point. @pikesley Do you think that restore pattern should default to False, to encourage people to be specific?

pikesley:
> I think that's a fair point. @pikesley Do you think that restore pattern should default to False, to encourage people to be specific?

I am genuinely ambivalent, what the Least Surprise we can offer here? `False` I guess?

benclifford:
I think False is the least surprise.

MatthewWilkes (approved):

PR #411, opened 2026-07-15
New approach to launch an app in the sim

### LINKED ISSUE #402: Sim not starting with override_app
**Describe the bug**

Previously I could start the sim with specified local app, e.g.
```
python run.py example.ExampleApp
```

Now I get 
```
Error: Module 'example' not found.
```

Git bisect takes me to 1c401fccd23fe738ec09caefbdf9dba036aa09d7 from #361

Version: current main v2.0.0-alpha.7

**To Reproduce**

In the `sim` directory run
`python run.py example.ExampleApp`

**Expected behavior**
Starts in the example app

**Screenshots**

**Additional context**
I am running the sim in a pixi environment. And I have had import related issues before #278

samtygier:
Launch by wrapping the Launcher with a subclass that selects a given item on startup.

Closes #402

# Description

The sim needs to do magical things to the import system in order to emulate running on the badge in micro python. The `override_app` options does more magic to launch an app by a its module path. This is all quite fragile and can break when import orders change elsewhere in the badge software (e.g. #278).

I think the approach of wrapping the Launcher class and calling `select_handler()` is more robust. If the sim can get as far as launching, then it should be able to launch any app that it had found.

I have not removed the old method, as maybe there are cases that it is useful. This change only touches `sim/run.py` and only does anything when the `--menu` option is used.

I have also added a bit of documentation.

### DIFF
--- a/sim/README.md
+++ b/sim/README.md
@@ -37,6 +37,23 @@ cd sim
 pipenv run python run.py
 ```
 
+You can launch an app at startup by its name in the menu:
+
+```
+pipenv run python run.py --menu "Example app"
+```
+
+or by its module path
+
+```
+pipenv run python run.py example.ExampleApp
+```
+
+Testing local apps
+---
+
+You can copy or symlink an app intp the `sim/apps/` directory. See `sim/apps/example` for an example.
+
 Known Issues
 ---
 
--- a/sim/run.py
+++ b/sim/run.py
@@ -169,6 +169,17 @@ def replace_launcher(module_name: str, class_name: str):
     import system.launcher.app
     system.launcher.app.Launcher = app_class
 
+def replace_launcher_wrap(menu_item: str):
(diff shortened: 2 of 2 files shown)


benclifford:
this is quite close to the autoexec.bat behaviour in #367 and one of the use cases of that was to support launching in the wasm emulator -- longer term might be interesting to understand how to make the autoexec.bat functionality also support this

samtygier:
I'd be interested in a shared generic approach to launching an app. It might also be useful in my File Browser app, e.g. to be able to launch an image viewer or text editor app (and give it a file path).

I don't think my change could hook into the existing `/autoexec.bat` support, as that is currently trying to read a file at the root of the host system (the sim does not have great isolation from the host). 

I'm open to ideas, but I think this is a simple solution for now, that avoids modifying anything it does not need to.

PR #425, opened 2026-07-22
Add RMT RX peripheral

MatthewWilkes:
Backports https://github.com/micropython/micropython/pull/18454 from MP 1.29 (or later)


ChrisDick (approved):
I think we still update to micropython 1.29 once it's released as some of the wifi changes look interesting.  pr18454 still hasn't been merged, so may be worth checking if anything changes when it does.

Issue #427, opened 2026-07-22
Confusing colours in the Settings app


grahamparks:
**Describe the bug**
It is extremely unintuitive what the currently focused item is in the Settings app, making it feel broken.

Having looked at the source code, item labels use orange for focused (and yellow for inactive), while button items use yellow for focused (and orange for inactive).

This bug also applies to other apps using the same components.

Version: 2.2.0, with 2026 front board.

**To Reproduce**
Open the Settings app, go up and down a few times, observe that there are often two items in orange and/or two items in yellow.

**Expected behavior**
There should be a single "focused" colour, and there should never be more than one item on screen in that colour.

Ideally, items that do nothing (some of the labels) would not use the focused colour.

(From an accessibility point of view, the focused item should also be indicated using something other than colour)

PR #429, opened 2026-07-22
Fixing colors of settings menu items to be less confusing.

grahamparks:
# Description

Changed the 2026 color scheme to be less confusing. Hopefully kept the 2024 color scheme the same.

See #427

### DIFF
--- a/modules/firmware_apps/settings_app.py
+++ b/modules/firmware_apps/settings_app.py
@@ -191,6 +191,8 @@ class SettingsApp(app.App):
                             return True
                         return False
 
+                    entry.button_handler = _button_event_pattern_toggle
+                    
                     entry = layout.ButtonDisplay(
                         "Next pattern", button_handler=_button_event_pattern_toggle
                     )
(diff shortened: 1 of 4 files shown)


modules/frontboards/twentysix.py
```diff
@@ -225,6 +225,8 @@ def ui_colors(self):
             "button_text": self.colors["black"],
             "active_button_background": self.colors["yellow"],
             "active_button_text": self.colors["black"],
+            "definition_label": colors["orange"],
+            "active_definition_label": colors["yellow"],
```
willselby: Is it intentional that they're flipped the opposite way round to 2024?

Also, these are missing the `self.` reference that the others have.

grahamparks: Yes, that's the fix! The 2024 design used four different colors (AFAICT, I didn't have one) so doesn't suffer from the same issue. 

The 2026 badge uses two colours but not consistently, hence needing them flipping here.

modules/firmware_apps/settings_app.py
```diff
@@ -191,6 +191,8 @@ async def _button_event_pattern_toggle(event):
                             return True
                         return False
 
+                    entry.button_handler = _button_event_pattern_toggle
```
ChrisDick, line 194: These get overwritten in the next instruction, so can be removed.

ChrisDick (changes requested):

PR #430, opened 2026-07-22
Setting to disable boop animation

Jaso5:
It boots a little faster when it's disabled, which makes frequent resetting of the device during development faster. I've added an entry inside SettingsApp to toggle this.

### DIFF
--- a/modules/firmware_apps/settings_app.py
+++ b/modules/firmware_apps/settings_app.py
@@ -239,6 +239,28 @@ class SettingsApp(app.App):
                     )
                     self.layout.items.append(entry)
 
+                if id == "enable_boot_animation":
+
+                    async def _button_event_mirror_toggle(event):
+                        if BUTTON_TYPES["CONFIRM"] in event.button:
+                            enable_boot_animation = settings.get(
+                                "enable_boot_animation"
+                            )
+
+                            # default will evaluate to False here
+                            enable_boot_animation = not enable_boot_animation
+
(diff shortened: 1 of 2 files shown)


modules/firmware_apps/settings_app.py
```diff
@@ -119,7 +119,11 @@ async def dev_mode(self, label, id, render_update):
         if self.devmode == 5:
             settings.set("developer", True)
             eventbus.emit(ShowNotificationEvent(message="You are now a developer"))
-        else:
+        elif self.devmode == 10:
+            eventbus.emit(
+                ShowNotificationEvent(message="You are now an anti-developer")
```
MatthewWilkes: 👀

Jaso5: I noticed you could overflow the devmode counter, so logically you become an anti-developer when it goes negative. Be careful you don't come in contact with a regular developer though...

hughrawlinson:
I like this overall, but the devmode counter thing is a little confusing, and an orthogonal change - I would hate for that change to block the addition of the setting, perhaps you could split the PRs?

MatthewWilkes:
Agreed. I think we'd be less inclined to merge the devmode one, to be honest. We should give people a way of getting out of dev mode, and I think that would probably involve just fixing the bounds check on that, rather than introducing an easter egg.

Jaso5:
I've removed the devmode commit.

hughrawlinson (approved):
Works perfectly on my badge

PR #439, opened 2026-07-25
Implement Standby mode and minor changes to Power Off

emericklaw:
# Description

This PR is in response to:

> Bob (he/him)
> If someone wants to open a PR so that pressing any button on the "you can now turn off your badge" screen  turns off the screen and LED power it'd definitely be appreciated!

The LCD backlight is connected on the module to 3v3 with no separate backlight pin. Looking at the schematic the 3v3 fed through to the screen is from the main 3v3 rail so cannot be toggled off.

I have come up with an alternative which clears the screen.

However, when testing this, I confirmed that due to `power.off()` disconnecting the battery, charging no longer works.

So I added a **Standby** mode which does the same as the current **Power Off** but does not call `power.off()`.

Also in response to:

>chrisdick
> it should really monitor the charge state, then turn off, will take a look at that pr after the weekend as I'm away again this weekend.

I modified the Power Off to monitor charge status before switching off.

### Main Menu

**Power Off** changed to **Power** since this now has a sub menu.

## Power Menu

Shows **Standby**, **Power Off** and **Storage**

## Standby

Disables LEDs, hexpansion slots and shows the Standby screen.

Pressing a key (not back/left) blanks the screen (backlight is still on).

## Exiting Standby

LEDs restored and hexpansion power.

## Power Off

Screen content can still be toggled off/on.

Initial screen shown when power off is choosen and USB is already plugged in.

Now after the charging has finished.

## Storage Mode

Since these badges can spend a lot of time not being used, I added a storage mode to get the battery to 3.8V before turning off.

### Voltage above 3.8V

Discharging mode, it will ask for USB to be disconnected if connected to allow discharging. The LEDs will go full brightness and slowly decrease in brightness when getting close to the 3.8V cut off to avoid voltage increasing when LEDs are switched off.

When 3.8V is reached the badge will turn off.

### Voltage below 3.8V

Charging mode, it will ask for USB to be connected if not already to allow charging. LEDs and hexpansion slots will be turned off.
The voltage is also adjusted due to the ~0.07V increase when being charged.

Once battery has reached 3.8V it will show the below stating it has finished charging.


nic0lette:
I don't know if this is helpful, but I did a bit of reading through the [GC9A01A datasheet](https://github.com/emfcamp/badge-2024-hardware/blob/main/datasheets/GC9A01A.pdf) and found there's a "Write CTRL Display" command, opcode 0x53, that isn't currently implemented (AFAIKT) in the `flow3r_bsp_gc9a01.*` files.

If implemented, it looks like the display can be instructed to turn the backlight off by clearing the 3rd bit (0x4) in the parameter.

emericklaw:
> I don't know if this is helpful, but I did a bit of reading through the [GC9A01A datasheet](https://github.com/emfcamp/badge-2024-hardware/blob/main/datasheets/GC9A01A.pdf) and found there's a "Write CTRL Display" command, opcode 0x53, that isn't currently implemented (AFAIKT) in the `flow3r_bsp_gc9a01.*` files.
> 
> If implemented, it looks like the display can be instructed to turn the backlight off by clearing the 3rd bit (0x4) in the parameter.

I did initially try this, but had no joy. I am pretty sure this is because the display board the badge uses connects backlight to 3v3.

emericklaw:
In response to 

> chrisdick
> it should really monitor the charge state, then turn off, will take a look at that pr after the weekend as I'm away again this weekend.

This PR changes the power off logic to monitor the charge status if already charging and only disconnects the battery when fully charged.

emericklaw:
Also added storage mode since the badges are not used by most people, this will hopefully save some spicy pillows...

I did go a bit overboard with the storage mode logic, it could be simplified if required.

modules/firmware_apps/power_app.py
```diff
@@ -0,0 +1,368 @@
...
+            (2, 12),
+            (2, 13),
+            (1, 8),
+            (1, 9),
+            (1, 10),
+            (1, 11),
+        }
+
+        for epin in HEXPANSION_POWER:
+            pin = ePin(epin, ePin.OUT)
```
ChrisDick, line 57: The Hexpansion supplies are turned off by setting the pin on as an output. 
The Hexpansion supplies are turned on by setting the pin as an input and shorting the detect pin to ground. 
Turning the pin off as an output forces the Hexpansion supply on and triggers the Hexpansion detect when no Hexpansion is present.  This is causing the port LEDs to come on when entering the app.

modules/firmware_apps/power_app.py
```diff
@@ -0,0 +1,368 @@
...
+            or BUTTON_TYPES["LEFT"] in event.button
+        ):
+            eventbus.remove(ButtonDownEvent, self._handle_buttondown, self)
+            eventbus.emit(PatternEnable())
+            self._hardware_power(on=True)
+            self.state = "menu"
+            self.screen_blank = False
+            self._prev_blank = False
+            self.menu = self._build_menu()
+            self.terminate()
```
ChrisDick, line 158: we shouldn't terminate here, it's valid to go back to the menu to go into another power mode and you've just rebuilt the menu.

modules/firmware_apps/power_app.py
```diff
@@ -0,0 +1,368 @@
...
+        else:
+            self._hardware_power(on=False)
+            self.state = "standby"
+        self._force_render = True
+        self._ignore_next_button = True
+        eventbus.on_async(ButtonDownEvent, self._handle_buttondown, self)
+
+    def _menu_back(self):
+        self.menu._cleanup()
+        self.minimise()
```
ChrisDick, line 134: should this be a terminate?

modules/firmware_apps/power_app.py
```diff
@@ -0,0 +1,368 @@
...
+
+    async def _handle_buttondown(self, event: ButtonDownEvent):
+        if self._ignore_next_button:
```
ChrisDick, line 137: is this in here to deal with an underlying bug?

modules/firmware_apps/power_app.py
```diff
@@ -0,0 +1,368 @@
...
+            self.menu.update(delta)
+            return True
+        # In poweroff/standby: only render when screen_blank changes
+        if self._force_render:
+            self._force_render = False
+            return True
+        if self._prev_blank != self.screen_blank:
+            self._prev_blank = self.screen_blank
+            return True
+        if self.state in ("standby", "charging_poweroff"):
```
ChrisDick, line 180: now that you've changed the poweroff state to check the charge state, is there much difference between standby and power off?

ChrisDick (changes requested):

PR #440, opened 2026-07-25
Feat/utc settings

sammachin:
This PR adds a new field to settings UTC Offset, the purpose is to allow the user to set their local timezone.

The badge gets its time from NTP, but `time.gmtime()` always returns UTC, there are several clock apps in the app store that all then add an offset to the time returned to display the "local" time, however these all seem to be hard coded to +1, basically British Summer Time, which is fine while at Camp but this means these apps will be wrong for half the year and anyone that takes a badge out of the UK will also have the wrong time!

Rather than doing this in my own app or adding my own setting it felt like it should be an OS level setting that apps can then call get `time.gmtime(now + settings.get("utc_offset", 0)`

Rather than using a complex tzinfo file and displaying locations like Europe/London etc I'm just using a simple scroll of all the possible UTC offsets (including the 30 and 45 min ones!) that then stores an int of seconds which can be applied to the time (see above)
This does mean that if you travel or when we switch from BST to GMT you need to adjust your offset but this feels like an ok compromise, being a common setting does mean that a developer could create a more polished app to manage setting based on location and DST changes but unless you're a seasoned jetsetter that takes your tildagon on many international flights this is probabbly overkill.

I've tested this setting on my own badge and also made a change to the https://github.com/tomsci/tildagon-clock to consume it (I'll PR that app)


ChrisDick (approved):
I think it's right to leave DST to the app developer, we can't support all variations of it in the firmware

